PR #1, opened 2013-08-19
populate with files from ros_industrial/abb

dpsolomon:
Also add meshes and URDFs for irb4400_l30


shaun-edwards:
@dpsolomon, my bad.  abb_experimental should just contain the new packages.  The new packages should follow this convention: http://ros.org/wiki/Industrial/Tutorials/SuggestedPackageLayoutNewRepositories.  You can also look at the fanuc repo to see an example.

dpsolomon:
OK, I will remove everything except the new files. Are you linking me to that page because I named something incorrectly? I believe all files are named consistently with the others.

PR #9, opened 2015-02-10
Added packages for ABB IRB 120 robotic manipulator

culletom:
As commented, this request is to add the abb_irb120_support and abb_irb120_moveit_config packages to the abb_experimental repository. Packages have been developed for ROS Hydro, and are structured with reference to the templates online.


abb_irb120_moveit_config/package.xml
```diff
@@ -0,0 +1,41 @@
+<package>
+  <name>abb_irb120_moveit_config</name>
+  <version>0.2.0</version>
+  <description>
+    <p>
+      MoveIt package for the ABB IRB 120.
+    </p>
+    <p>
+      An automatically generated package with all the configuration and launch
+      files for using the ABB IRB 2400 with the MoveIt Motion Planning
```
shaun-edwards: Reference to IRB 2400 (copy and paste error?)

shaun-edwards:
There appears to be variants of the IRB 120 (see [here](http://new.abb.com/products/robotics/industrial-robots/irb-120/irb-120-data)).  What variant does the PR cover.  Also, see issue ros-industrial/abb#47 for a description of variants.  See the fanuc [LRMate200ic](https://github.com/ros-industrial/fanuc/tree/hydro-devel/fanuc_lrmate200ic_support) for an example of a package with variants.

shaun-edwards:
@Levi-Armstrong, please review as well.

abb_irb120_moveit_config/package.xml
culletom: Yes just a copy and paste error, apologies

abb_irb120_moveit_config/config/joint_limits.yaml
```diff
@@ -0,0 +1,31 @@
+joint_limits:
+  joint_1:
+    has_velocity_limits: true
+    max_velocity: 4.36332
+    has_acceleration_limits: false
+    max_acceleration: 0
```
gavanderhoorn: Missing acceleration limits. I've seen strange planning results from MoveIt without them, but I don't know what the current state of this is.

I've always just either set them to the real limits (if available), or used the 'old way' of setting them to `1/5 * joint_vel_limit`.

abb_irb120_moveit_config/config/kinematics.yaml
```diff
@@ -0,0 +1,6 @@
+manipulator:
+  #kinematics_solver: abb_irb120_manipulator_kinematics/IKFastKinematicsPlugin
+  kinematics_solver: kdl_kinematics_plugin/KDLKinematicsPlugin
+  kinematics_solver_search_resolution: 0.005
+  kinematics_solver_timeout: 0.005
+  kinematics_solver_attempts: 3
```
gavanderhoorn, line 5: (optional) this moveit cfg has no ikfast plugin. We may want to log an issue for that.

abb_irb120_moveit_config/package.xml
gavanderhoorn: Note also the 2400 references later on in this file.

abb_irb120_moveit_config/package.xml
```diff
@@ -0,0 +1,41 @@
...
+    </p>
+  </description>
+  <author email="assistant@moveit.ros.org">MoveIt Setup Assistant</author>
+  <maintainer email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</maintainer>
+
+  <license>BSD</license>
+
+  <url type="website">http://moveit.ros.org/</url>
+  <url type="bugtracker">https://github.com/ros-industrial/abb/issues</url>
+  <url type="repository">https://github.com/ros-industrial/abb</url>
```
gavanderhoorn: package is in `abb_experimental`, not in `abb`.

abb_irb120_moveit_config/package.xml
```diff
@@ -0,0 +1,41 @@
...
+  <run_depend>moveit_ros_visualization</run_depend>
+  <run_depend>joint_state_publisher</run_depend>
+  <run_depend>robot_state_publisher</run_depend>
+  <run_depend>moveit_ros_move_group</run_depend>
+
+  <run_depend>xacro</run_depend>
+
+  <run_depend>industrial_robot_simulator</run_depend>
+  <run_depend>abb_irb2400_support</run_depend>
+  <run_depend>abb_irb2400_moveit_plugins</run_depend>
```
gavanderhoorn: without an ikfast plugin for the irb120, this dependency is unneeded (after resolving the 2400 copy/pasta).

abb_irb120_support/launch/load_irb120.launch
```diff
@@ -0,0 +1,3 @@
+<launch>
+	<param name="robot_description" command="$(find xacro)/xacro.py '$(find abb_irb120_support)/urdf/irb120.xacro'" />
```
gavanderhoorn: indent: use spaces (2 per level).

abb_irb120_support/package.xml
```diff
@@ -0,0 +1,41 @@
+<package>
+  <name>abb_irb120_support</name>
+  <version>0.2.0</version>
+  <description>
+    <p>
+      ROS-Industrial support for the ABB IRB 120
```
gavanderhoorn: Missing dot at end of line.

abb_irb120_support/package.xml
```diff
@@ -0,0 +1,41 @@
...
+    </p>
+    <p>
+      This package contains configuration data, 3D models and launch files
+      for ABB IRB 120 manipulator. This currently includes the base model.
```
gavanderhoorn: Perhaps add 'only': "This currently includes the base model only.".

abb_irb120_support/package.xml
```diff
@@ -0,0 +1,41 @@
...
+    </p>
+    <p>
+      Joint limits and max joint velocities are based on the information in
+      the ABB data sheets.  All URDFs / XACROs are based on the
```
gavanderhoorn: Which data sheet in particular did you use? Ideally we'd provide a link (or document nr), so users can look it up and check their own robot.

abb_irb120_support/package.xml
```diff
@@ -0,0 +1,41 @@
...
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</author>
+  <maintainer email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</maintainer>
+  <license>Apache2</license>
+
+  <url type="website">http://moveit.ros.org/</url>
+  <url type="bugtracker">https://github.com/ros-industrial/abb/issues</url>
+  <url type="repository">https://github.com/ros-industrial/abb</url>
```
gavanderhoorn: again `abb` vs `abb_experimental`.

abb_irb120_support/package.xml
```diff
@@ -0,0 +1,41 @@
...
+      Before using any of the configuration files and / or meshes included
+      in this package, be sure to check they are correct for the particular
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</author>
+  <maintainer email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</maintainer>
+  <license>Apache2</license>
+
+  <url type="website">http://moveit.ros.org/</url>
```
gavanderhoorn: This is a support package, not a MoveIt package, right? `website` should probably be something like `http://wiki.ros.org/abb_irb120_support`.

abb_irb120_support/urdf/irb120.xacro
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" ?>
+
+<robot name="abb_irb120" xmlns:xacro="http://ros.org/wiki/xacro">
+	<xacro:include filename="$(find abb_irb120_support)/urdf/irb120_macro.xacro"/>
+	<xacro:abb_irb120 prefix=""/>
```
gavanderhoorn: indent (see earlier comment).

abb_irb120_support/urdf/irb120_macro.xacro
```diff
@@ -0,0 +1,195 @@
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+<xacro:macro name="abb_irb120" params="prefix">
+	<!-- link list -->
+	<link name="${prefix}base_link">
+		<visual>
+			<origin xyz="0 0 0" rpy="0 0 0"/>
+			<geometry>
+				<mesh filename="package://abb_irb120_support/meshes/irb120/visual/base_link.STL"/>
```
gavanderhoorn: I know the solidworks to urdf exporter does this by default (I don' t know if you've used it), but we try to keep all file extensions lower case.

abb_irb120_support/urdf/irb120_macro.xacro
```diff
@@ -0,0 +1,195 @@
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+<xacro:macro name="abb_irb120" params="prefix">
+	<!-- link list -->
```
gavanderhoorn: Please change the indentation to two spaces per level. I know the other xacros don't do this (yet), but we should try to do this in the new ones. I've used [expand](http://linux.die.net/man/1/expand) for this earlier.

abb_irb120_support/urdf/irb120_macro.xacro
```diff
@@ -0,0 +1,195 @@
...
+			<material name="black">
+			    <color rgba="0 0 0 1"/>
+			</material>	
+		</visual>
+		<collision>
+			<origin xyz="0 0 0" rpy="0 0 0"/>
+			<geometry>
+				<mesh filename="package://abb_irb120_support/meshes/irb120/collision/link_6.STL"/>
+			</geometry>
+            <material name="yellow"/>
```
gavanderhoorn: this seems like it's not correctly indented (also for other material elements earlier)?

abb_irb120_support/urdf/irb120_macro.xacro
```diff
@@ -0,0 +1,195 @@
...
+
+	<!-- joint list -->
+	<joint name="${prefix}joint_1" type="revolute">
+		<origin rpy="0 0 0" xyz="0 0 0"/>
+		<parent link="${prefix}base_link"/>
+		<child link="${prefix}link_1"/>
+		<limit effort="0" lower="-2.87979" upper="2.87979" velocity="4.36332"/>
+		<axis xyz="0 0 1"/>
+        <calibration rising="0" falling="0"/>
+		<dynamics damping="0" friction="0"/>
```
gavanderhoorn: Not sure whether we want to keep these elements in a basic urdf. Do you have any specific use for them?

@shaun-edwards ?

gavanderhoorn:
@shaun-edwards: this appears to model the 'base' IRB 120, which is also mentioned in the support package manifest (see [datasheet](http://www05.abb.com/global/scot/scot241.nsf/veritydisplay/3bd625bab3c7cae1c1257a0800495fac/$file/ROB0149EN_D_LR.pdf)). So I think the use of `abb_irb120` is correct, and does not need any further suffixes.

abb_irb120_moveit_config/config/joint_limits.yaml
culletom: I have contacted ABB about getting real acceleration limits, but for the minute I have used the 'old way' of setting them as above.

abb_irb120_support/urdf/irb120_macro.xacro
culletom: I can remove these if needs be, I was just following the IRB 2400 description

gavanderhoorn:
@culletom: it seems the PR now introduces meshes with both lowercase and uppercase extensions, in both `collision` and the `visual` subdirs?

culletom:
Yes just to confirm that this is for the base IRB 120 model. As far as I can tell, the only difference between the two variants is their joint speed limits, so I will create the required files for the IRB 120T model as well, and will arrange with reference to the [fanuc LRMate200ic](https://github.com/ros-industrial/fanuc/tree/hydro-devel/fanuc_lrmate200ic_support) package

abb_irb120_moveit_config/package.xml
```diff
@@ -0,0 +1,40 @@
...
+      Framework.
+    </p>
+  </description>
+  <author email="assistant@moveit.ros.org">MoveIt Setup Assistant</author>
+  <maintainer email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</maintainer>
+
+  <license>BSD</license>
+
+  <url type="website">http://moveit.ros.org/</url>
+  <url type="bugtracker">https://github.com/ros-industrial/abb/issues</url>
```
gavanderhoorn: I'm not sure, but perhaps this should point to the issue tracker of the `abb_experimental` repository? All other experimental packages do so, I believe.

abb_irb120t_moveit_config/package.xml
```diff
@@ -0,0 +1,40 @@
...
+      Framework.
+    </p>
+  </description>
+  <author email="assistant@moveit.ros.org">MoveIt Setup Assistant</author>
+  <maintainer email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</maintainer>
+
+  <license>BSD</license>
+
+  <url type="website">http://moveit.ros.org/</url>
+  <url type="bugtracker">https://github.com/ros-industrial/abb/issues</url>
```
gavanderhoorn: See earlier comment.

abb_irb120_support/package.xml
```diff
@@ -0,0 +1,41 @@
...
+      in this package, be sure to check they are correct for the particular
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</author>
+  <maintainer email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</maintainer>
+  <license>Apache2</license>
+
+  <url type="website">http://wiki.ros.org/abb</url>
+  <url type="bugtracker">https://github.com/ros-industrial/abb/issues</url>
```
gavanderhoorn: See earlier comment.

gavanderhoorn:
@culletom: this looks good (btw: reformatting of the MoveIt pkg launch files was not necessary: they are autogenerated like that, and it is too much work to keep doing that every time something changes).

The `visual` meshes could probably be decimated a bit, but that is not important for now (we can log an issue).

I'd still rather remove the `dynamics` and `calibration` elements from the xacros (as they are never used), but let's hear some other opinions. @shaun-edwards, @Levi-Armstrong?

Levi-Armstrong:
I agree, if they are never used it is probably best to remove the dynamics and calibration elements from xacros.

gavanderhoorn:
Well, they can be used, but only in simulation contexts, which none of the 'basic' ros-i urdfs currently support / are setup for.

Personally I like to separate responsibilities as much as possible, so if the simulation properties (elements) could be added in some other way, that would have my preference. If that is not possible however, then we could leave them in.

culletom:
Ok those elements are removed from the xacros, hopefully the packages are now taking shape. Apologies for so many mistakes, I'll make sure to keep the above in mind when creating ones in the future.  
As an aside, is there a reason why current ABB packages don't support simulation? Just asking as I'm currently working on Gazebo packages for the IRB 120, so could potentially upload something if it would be useful.

gavanderhoorn:
@culletom: no need to apologise for anything. I can be a difficult person when it comes to these reviews, and it's not like we've published these conventions anywhere (or at least haven't covered all the details).

Thank _you_ for taking the time to contribute these packages.

> As an aside, is there a reason why current ABB packages don't support simulation? Just asking as I'm currently working on Gazebo packages for the IRB 120, so could potentially upload something if it would be useful. 

The only reason (if we can call it that) is that no one has really had a need for Gazebo compatibility until now, I think. So: yes, PRs adding Gazebo packages for the various robot models in the ABB support packages would be very much welcomed.

Levi-Armstrong:
@gavanderhoorn: I believe all requested changes have been implemented, do you agree?

gavanderhoorn:
I think so, yes. We can merge this and fix anything we missed in later PRs.

culletom:
That is great to hear, I appreciate all the feedback. The only thing that has been brought up that hasn't been addressed is the decimation of the visual links, but as suggested this can be logged as an issue - I will try do this over the next few days, as well as cleaning up the gazebo packages that I am currently working on.

gavanderhoorn:
Hm, seems I missed one additional thing. Not that important right now (as all ABB support pkgs have similar issues), but the translations between the joints in the urdfs seem 'odd'. While the final transform to `tool0` is probably correct, I wonder whether fi the frame for `joint_4` is at the right location (`<origin rpy="0 0 0" xyz="0.1496 -0.00475 0.07"`).

I'll log an issue for this.

gavanderhoorn:
But really: thanks for contributing @culletom, very much appreciated.

PR #12, opened 2015-02-19
Intial commit for ABB irb_1600 support package

asct:


abb_irb1600_support/CHANGELOG.rst
```diff
@@ -0,0 +1,13 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
gavanderhoorn, line 1: This seems strange: has this pkg been released already?

abb_irb1600_support/config/joint_names_irb1600.yaml
```diff
@@ -0,0 +1 @@
+controller_joint_names: ['joint_1', 'joint_2', 'joint_3', 'joint_4', 'joint_5', 'joint_6']
```
gavanderhoorn, line 1: The name of this file should actually include the variant specific suffixes ('12'?).

gavanderhoorn:
Some comments. I've opted to not do them inline this time:
- meshes have uppercase file extensions (keep everything lower case)
- multiple references to IRB 2400 (launch files)
- variant suffix (`_12`) seems not specific enough. Assuming `12` is actually '1.2', is this the `6 / 1.2` or `10 / 1.2` (from [here](http://www08.abb.com/global/scot/scot241.nsf/veritydisplay/3b0491a94bd700a248257c71004ef393/$file/PR10282EN_R8.pdf))?
- variant suffix should be used throughout (file names, content, directory names)
- if meshes are usable for all models, `meshes` subdir should not include variant suffix (or at least split out the common ones)
- what is the 'base model' of the IRB 1600 (pkg manifest)?
- which data sheets were used, specifically (url) (pkg manifest)?
- the package is in `abb_experimental`, not in `abb` (pkg manifest)
- urdf and xacro seem out-of-sync: orientation of `tool0` is `0 1.57079 0` in urdf, but `0 .57079 0` in xacro?
- indentation in xacros should be two spaces per level
- xml elements in xacro should be as much as possible formatted as single lines (origin elements fi), unless they have child elements (easily solved with something like [xmlprettyprint.com](http://xmlprettyprint.com/)
- `calibration` and `dynamics` elements should be removed (or updated, `0` seems like default values)

culletom:
Just FYI, I remember coming across a pull request for this model before, see [here](https://github.com/ros-industrial/abb/pull/31). From the discussion it seems like it was meant to be pulled into abb_experimental, but that didn't end up happening. Still might be worth referencing though.

gavanderhoorn:
On 20/02/15 13:05, Mark Culleton wrote:

> Just FYI, I remember coming across a pull request for this model before, see [here](https://github.com/ros-industrial/abb/pull/31). From the discussion it seems like it was meant to be pulled into abb_experimental, but that didn't end up happening. Still might be worth referencing though.

yep, true.

If this PR uses any of the files from the old one, it would be nice to 
mention @Samsagax.

Samsagax:
I think they are, but for another branch I pushed in my personal fork. Is this by any chance the one asked for by Paul?

Here is my branch if you want to compare: https://github.com/Samsagax/abb/tree/hydro-devel-add-irb-1600

(or you can just pull mine, it has all that's needed for the 1600-1.2 version)

asct:
Joaquín,

This is the same URDF that Paul requested I just created the support package for it and I’m going to create another one for the irb_1600/1.45 

Ahmed Abdelmawgoud 

Research Engineer 

Robotics and Automation Engineering 

Office: (210)522-6027

Cell:    (210)284-6664

 mailto:aabdelmawgoud@swri.org aabdelmawgoud@swri.org

robotics.swri.org

Samsagax:
Actually it's great. I was about to make it too, but I switched jobs and no longer had a physical manipulator to test it. This raises the question (again I think?) about naming convention for different variations of same robot model. We should have:
- irb_1600_6_12
- irb_1600_10_12
- irb_1600_6_145
- irb_1600_10_145

I think there are some common meshes and each variant uses a particular combination of those (links 2, 3, 4 are the variants from the top of my head). So all meshes should be in a common directory and suffix the variants. Don't you think?

gavanderhoorn:
@Samsagax: that would indeed be a possible naming scheme.

As to the meshes: any meshes that can be shared between variants, should end up in a common directory (`irb1600`). The variants can then store any specific meshes in variant specific directories (ie `irb1600_6_12` etc).

We need to make sure that variant naming is very precise, so as to not to make the same mistake as with some of the other pkgs in the `abb` repository (see ros-industrial/abb#47).

Levi-Armstrong:
I reviewed most of the ABB robots and they can be described by Model, Payload and Reach as suggested by @Samsagax. The possible issue could be the removal of the decimal in the case where the robot has a payload of 1.2 and 12 which is unlikely since most industrial robot payloads are advertised as integers. 

A suggested fix could be replace the decimal with a letter like p which would 1p2 and 12 which I don't particular like but it would eliminate the possible issue.

gavanderhoorn:
Apart from the naming (for which we have at least _some_ guidelines [here](http://wiki.ros.org/Industrial/Tutorials/SuggestedPackageLayoutNewRepositories#General)), the other issues with this PR need to be addressed before it can be merged.

AustinDeric:
@asct are you still working on this?  If not, i am going to make the changes, close this PR and open a new one.

Thank you!

AustinDeric:
I talked to @asct and he said i could close this PR and open a new one for the fixes.

Issue #10, opened 2015-02-19
irb120: meshes can probably be decimated


gavanderhoorn:
Not really high-priority, but some of the visual meshes can probably be decimated a bit. That shouldn't have an effect on visual quality.

PR #13, opened 2015-02-25
Fix #11, recreated stl and urdf to remove odd transformation

### LINKED ISSUE #11: irb120: 'odd' transforms between joints
Some of the joints in the IRB 120 (T) urdfs have transforms where a translation is defined over more than 1 axis, and for which these translations are sometimes very small (< 1mm fi). While this is perfectly legal, most industrial robots have relatively simple kinematic chains, where axes are only offset in one dimension.

The `base->tool0` transform is probably ok, but transforming points into any of the intermediate frames could give incorrect results (when compared to the industrial controller).

This may be a consequence of using the SolidWorks to URDF converter, as it has a tendency to place joints at strange locations.

Levi-Armstrong:


gavanderhoorn:
Looking good. This solves #11 as far as I can tell.

(only nitpick is rotation necessary for the `base_link` mesh, but the s2u exporter always does that to me as well).

Levi-Armstrong:
Yea, it was bothering me too; I went ahead and fixed the unnecessary
rotation.

On Wed, Feb 25, 2015 at 1:55 AM, G.A. vd. Hoorn notifications@github.com
wrote:

> Looking good. This solves #11
> https://github.com/ros-industrial/abb_experimental/issues/11 as far as
> I can tell.
> 
> (only nitpick is rotation necessary for the base_link mesh, but the s2u
> exporter always does that to me as well).
> 
> —
> https://github.com/ros-industrial/abb_experimental/pull/13#issuecomment-75919522
> .

## 

Levi Armstrong

gavanderhoorn:
Nice. If I may suggest: just squash this into one commit.

Levi-Armstrong:
@gavanderhoorn, Did I performed the squash correctly; it looks correct but
want to make sure before it is merged?

On Wed, Feb 25, 2015 at 8:48 AM, G.A. vd. Hoorn notifications@github.com
wrote:

> Nice. If I may suggest: just squash this into one commit.
> 
> —
> https://github.com/ros-industrial/abb_experimental/pull/13#issuecomment-75972115
> .

## 

Levi Armstrong

gavanderhoorn:
Looks ok to me.

culletom:
@Levi-Armstrong thanks for the fixes. Could I just ask how you went about
this, so that I can improve any future commits? Also, does this address
issue #10 <https://github.com/ros-industrial/abb_experimental/issues/10> as
well?

On 26 February 2015 at 07:25, G.A. vd. Hoorn <notifications@github.com>
wrote:

> Looks ok to me.
>
> —
> <https://github.com/ros-industrial/abb_experimental/pull/13#issuecomment-76133522>
> .
>



-- 
*Mark Culleton*

PhD Candidate
Robotics Research Group
Department of Mechanical & Manufacturing Engineering
Trinity College Dublin
Funding Body: Irish Research Council
E: culletom@tcd.ie
P: (+353) 86 1772095

culletom:
Tried editing my previous comment, but for some reason the pull request page keeps crashing so can't get a chance. Realized after I posted that the visual meshes are similar to before, but perhaps they could be decimated before this commit? I also noticed the number of faces in the collision meshes have increased, which is slowing down the motion planner in moveit, so perhaps they could be reduced too? If not then I can do both of these in a separate pull request.

gavanderhoorn:
@culletom: lets leave the decimation for another PR. It's not terribly important, and I'd rather have separate PRs.

> I also noticed the number of faces in the collision meshes have increased, which is slowing down the motion planner in moveit, so perhaps they could be reduced too?

Are you really seeing an impact on planning time with these meshes? I just checked and they are around 2K faces for `link_2.stl` fi, which doesn't seem excessive (but lower is always good, as long as it still faithfully represents the geometry).

culletom:
Ok that is fair enough, I'll submit a separate PR at a later stage. Yea
there is a notable difference in planning times for me, but I guess it can
be put down to my computer's specs.

Levi-Armstrong:
@culletom, No problem, I start by finding the product specification
document for the robot which contains a diagram of the kinematics. I use
this as a guide to layout each joint frame such that each transformation
occurs along one axis and are on the same plane, which is usually possible
with most industrial robots. This also allows for someone to easily compare
the urdf to the product specification to verify the kinematics are
correctly defined. In this case I used SolidWorks to accomplish this task.
I define the axis for joint 2, 3, & 5 using a geometric feature of the CAD
model followed by finding the intersection with the axis and the xz-plane
defining the origin of the frame for joint 2, 3, & 5. I then check the
measurement from the base frame to the these frames and it should match the
diagram, if not then the geometric feature used to create the axis was not
the correct surface to use or the assembly was created incorrectly. Joint 4
is similar but I find the intersection between its axis and the yz-plane to
define the origin of joint 4 and for joint 6, I find the intersection
between axis 4 and the plane of the mount flange to define the origin of
joint 6.  Hope this helps.

On Thu, Feb 26, 2015 at 3:53 AM, Mark Culleton notifications@github.com
wrote:

> @Levi-Armstrong thanks for the fixes. Could I just ask how you went about
> this, so that I can improve any future commits? Also, does this address
> issue #10 https://github.com/ros-industrial/abb_experimental/issues/10
> as
> well?
> 
> On 26 February 2015 at 07:25, G.A. vd. Hoorn notifications@github.com
> wrote:
> 
> > Looks ok to me.
> > 
> > —
> > Reply to this email directly or view it on GitHub
> > <
> > https://github.com/ros-industrial/abb_experimental/pull/13#issuecomment-76133522
> > 
> > .
> 
> ## 
> 
> _Mark Culleton_
> 
> PhD Candidate
> Robotics Research Group
> Department of Mechanical & Manufacturing Engineering
> Trinity College Dublin
> Funding Body: Irish Research Council
> E: culletom@tcd.ie
> P: (+353) 86 1772095
> 
> —
> https://github.com/ros-industrial/abb_experimental/pull/13#issuecomment-76150301
> .

## 

Levi Armstrong

culletom:
@Levi-Armstrong thanks that will be very helpful next time I'm creating a
urdf

gavanderhoorn:
Note btw that the translations shown in the spec sheets don't always correspond to the D-H data (or similar) the controller is using. This means that transforms to intermediate frames may still no be completely identical to the data the controller would come up with (if it has that capability).

PR #15, opened 2015-02-27
Fix #10: Decimated collision meshes to ~100 faces

### LINKED ISSUE #10: irb120: meshes can probably be decimated

Levi-Armstrong:


culletom:
Just in regards to this, issue #10 relates to the visual meshes rather than the collision meshes. However the collision meshes also needed to be decimated (for my laptop anyways) so perhaps both can be brought in in the PR?

Levi-Armstrong:
@culletom: You are correct, I have included the decimated visual meshes.

gavanderhoorn:
@Levi-Armstrong: could you update your commit msg to include the name of the pkgs that you updated the meshes for? Something like `irb120: ..` would be nice. That way the history nicely shows what is going on, without having to look at the contents of commits.

gavanderhoorn:
@culletom: could you perhaps check whether the meshes still resemble what you'd expect? I feel you are in a good position to judge, seeing as you contributed the original set(s).

Levi-Armstrong:
@gavanderhoorn: No problem, I also went ahead and combined the two commits in to one along with updating the commit message.

culletom:
Just had a look at these and runs much smoother on my laptop. However, I would suggest reverting the base_link collision mesh, as it is causing collisions in Gazebo which is causing the model to vibrate.

Levi-Armstrong:
@culletom: Is it showing a collision with link_1; if so is it possible to not check for collision between base_link and link_1 as in Moveit?

culletom:
@Levi-Armstrong No its to do with the base_link colliding with the gazebo
world. I have tried just loading the base_link into gazebo, but the same
problem exists - it seems that the decimated base_link collision mesh
extends below the gazebo world plane.

Levi-Armstrong:
@culletom do you mind checking how much the mesh needs to be translated up to alleviate the problem?

gavanderhoorn:
On 19-3-2015 16:14, Levi Armstrong wrote:

> @culletom do you mind checking how much the mesh needs to be translated up to alleviate the problem?

This sounds a bit like an xy-problem.

Perhaps we need to first figure out how simply creating a convex hull
could suddenly cause a mesh to change in such a way that there are now
collision where there weren't.

Levi-Armstrong:
The issue was not with creating the convex hull but with the decimation process. Since I had decimated  it so much certain features in the detailed mesh was protruding the collision mesh. This required me to scale the collision mesh slightly to enclose the detailed mesh.  I changed the way it was scaled so it should be fixed. @culletom see if the new mesh fixes the issue.

culletom:
@Levi-Armstrong The new mesh is much better, but sill causes slight
vibrations. I checked and a mesh translation of 0.3 mm in the z-direction
alleviates the problem, so the decimated mesh must still larger by this
distance.

gavanderhoorn:
@Levi-Armstrong: you're right, and I was confused, but even (or especially) decimation should never result in changes to the geometry so severe that they can result in the issue @culletom reported. Especially not convex hulls, as they are supposed to be convex approximations.

Levi-Armstrong:
1) The visual starts out with the base around -9e-10m below the xy-plane. 
2) After convex hull operation it is -1e-9m 
3) After the decimation it is -2.86e-4m
4) After final convex hull operation -2.86e-4m = -2.86mm

This explains why the 0.3 mm z-direction shift solved the issue. I am curious have you tired using the visual mesh, if so does the issue still exist?

Levi-Armstrong:
I found the option in the decimation GUI that caused the issue, by default it has Optimal position of simplified vertices checked.  When unchecked:
3) After the decimation it is -1e-9m
4) After final convex hull operation -1e-9m = -1e-6mm

gavanderhoorn:
Ok, great. @culletom: does that fix the issues you were having?

culletom:
Yes this fixes the problem, thanks for sorting it out! Everything else
seems good, so I'm happy for this to go through.

PR #18, opened 2015-02-27
Introduction of Gazebo Support

culletom:
Wanted to discuss this before going any further. I've looked through the ros_industrial repository and there are are very few gazebo packages (only in the [universal_robot](https://github.com/ros-industrial/universal_robot) repository as far as I could see) so it would be worth discussing to try come up with a standard package format. 

As can be seen, I've created seperate urdf files for Gazebo, as I felt it was best not to complicate the original urdfs since they are used by other programs. The irb120_gazebo_macro file contains all the required tags for gazebo. This differs to some of the gazebo tutorials, but ensures all the tags use the same prefix parameter. 

The masses and inertial values for each link were calculated in solidworks. Adding the mass of each link gives a total of 18kg, which is lower than the robots actual value of 25kg. I contacted ABB about this, but the factory doesn't provide more accurate values for each link. They did say that solidworks values are accurate, but that they don't include the robot cable harness as the mass distribution is difficult to calculate. I updated the inertia tensors and cog origins to account for the new joint translations, but it might be worth double checking these values. The inertia tensor for link 6 seemed was causing problems in Gazebo (values were too small) so I ended up using a simplified matrix for this link.

I have included the dynamics tag in the urdf, as it is used by Gazebo. However, following the universal_robot package, I let both damping and friction equal 0. I also increased the effort limits for each joint - these values were somewhat arbitrarily chosen. 

The gazebo ros_control controllers have been setup to receive joint trajectory commands. I've tuned their PID values, but I'm sure they are not perfect. However they do allow for moveit! control. I've outlined how to do this in the readme file. To note: I've kept this as three separate launch files, as it allows the controllers to be shut down cleanly prior to exiting gazebo [discussed here](https://github.com/ros-controls/ros_control/issues/195). 

A final comment is that I had to slightly modify the moveit_planning_execution launch file in each of the moveit_config packages. I know this is disruptive, so that is why I wanted to discuss this before going any further. The advantage of the way I have done it is that moveit can be launched using the moveit_planning_execution launch file regardless of whether gazebo is being used (unlike the universal_robot package). But let me know what yous think.

### DIFF
--- a/abb_irb120_gazebo/CMakeLists.txt
+++ b/abb_irb120_gazebo/CMakeLists.txt
@@ -1,0 +1,13 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(abb_irb120_gazebo)
+
+find_package(catkin REQUIRED)
+
+catkin_package()
+
+foreach(dir config launch urdf)
+   install(DIRECTORY ${dir}/
+      DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/${dir})
+endforeach()
+
+install(FILES readme.md DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
--- a/abb_irb120_gazebo/config/irb120_arm_controller.yaml
+++ b/abb_irb120_gazebo/config/irb120_arm_controller.yaml
@@ -1,0 +1,21 @@
+arm_controller:
+  type: position_controllers/JointTrajectoryController
+  joints:
+     - joint_1
+     - joint_2
+     - joint_3
+     - joint_4
+     - joint_5
+     - joint_6
+  constraints:
+      goal_time: 0.6
+      stopped_velocity_tolerance: 0.05
+      joint_1: {trajectory: 0.1, goal: 0.1}
+      joint_2: {trajectory: 0.1, goal: 0.1}
+      joint_3: {trajectory: 0.1, goal: 0.1}
+      joint_4: {trajectory: 0.1, goal: 0.1}
+      joint_5: {trajectory: 0.1, goal: 0.1}
+      joint_6: {trajectory: 0.1, goal: 0.1}
+  stop_trajectory_duration: 0.5
+  state_publish_rate:  25
+  action_monitor_rate: 10
--- a/abb_irb120_gazebo/config/joint_state_controller.yaml
+++ b/abb_irb120_gazebo/config/joint_state_controller.yaml
@@ -1,0 +1,3 @@
+joint_state_controller:
+    type: joint_state_controller/JointStateController
+    publish_rate: 50  
--- a/abb_irb120_gazebo/launch/irb120_control.launch
+++ b/abb_irb120_gazebo/launch/irb120_control.launch
@@ -1,0 +1,11 @@
+<launch>
+
+  <!-- load the joint state controller -->
+  <rosparam file="$(find abb_irb120_gazebo)/config/joint_state_controller.yaml" command="load" />
+  <node name="joint_state_controller_spawner" pkg="controller_manager" type="controller_manager" args="spawn joint_state_controller" />
+
+  <!-- load the arm controller -->
+  <rosparam file="$(find abb_irb120_gazebo)/config/irb120_arm_controller.yaml" command="load" />
+  <node name="abb_irb120_controller_spawner" pkg="controller_manager" type="controller_manager" args="spawn arm_controller" />
+  
+</launch>
--- a/abb_irb120_gazebo/launch/irb120_gazebo.launch
+++ b/abb_irb120_gazebo/launch/irb120_gazebo.launch
@@ -1,0 +1,30 @@
+<launch>
+	<arg name="paused" default="false"/>
+
+  <!-- remap topics to conform to ROS-I specifications -->
+  <remap from="/arm_controller/follow_joint_trajectory" to="/joint_trajectory_action" />
+  <remap from="/arm_controller/state" to="/feedback_states" />
+  <remap from="/arm_controller/command" to="/joint_path_command"/>
+
+  <!-- startup simulated world -->
+  <include file="$(find gazebo_ros)/launch/empty_world.launch">
+    <arg name="world_name" value="worlds/empty.world"/>
+    <arg name="gui" value="true"/>
+    <arg name="paused" value="$(arg paused)"/>
+  </include>
+
+  <!-- urdf xml robot description loaded on the Parameter Server, converting the xacro into a proper urdf file-->
+  <include file="$(find abb_irb120_gazebo)/launch/load_irb120.launch" />
+
+  <!-- push robot_description to factory and spawn robot in gazebo -->
+  <node name="abb_irb120_spawn" pkg="gazebo_ros" type="spawn_model" output="screen" args="-urdf -param robot_description -model abb_irb120" />
+
+    <!-- convert joint states to TF transforms for rviz, etc -->
+  <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher"
+  output="screen">
+  </node>
+
+  <!-- init and start Gazebo ros_control interface -->
+  <include file="$(find abb_irb120_gazebo)/launch/irb120_control.launch"/>
+
+</launch>
--- a/abb_irb120_gazebo/launch/irb120t_gazebo.launch
+++ b/abb_irb120_gazebo/launch/irb120t_gazebo.launch
@@ -1,0 +1,30 @@
+<launch>
+	<arg name="paused" default="false"/>
+
+  <!-- remap topics to conform to ROS-I specifications -->
+  <remap from="/arm_controller/follow_joint_trajectory" to="/joint_trajectory_action" />
+  <remap from="/arm_controller/state" to="/feedback_states" />
+  <remap from="/arm_controller/command" to="/joint_path_command"/>
+
+  <!-- startup simulated world -->
+  <include file="$(find gazebo_ros)/launch/empty_world.launch">
+    <arg name="world_name" value="worlds/empty.world"/>
+    <arg name="gui" value="true"/>
+    <arg name="paused" value="$(arg paused)"/>
+  </include>
+
+  <!-- urdf xml robot description loaded on the Parameter Server, converting the xacro into a proper urdf file-->
+  <include file="$(find abb_irb120_gazebo)/launch/load_irb120t.launch" />
+
+  <!-- push robot_description to factory and spawn robot in gazebo -->
+  <node name="abb_irb120t_spawn" pkg="gazebo_ros" type="spawn_model" output="screen" args="-urdf -param robot_description -model abb_irb120t" />
+
+    <!-- convert joint states to TF transforms for rviz, etc -->
+  <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher"
+  output="screen">
+  </node>
+
+  <!-- init and start Gazebo ros_control interface -->
+  <include file="$(find abb_irb120_gazebo)/launch/irb120_control.launch"/>
+  
+</launch>
--- a/abb_irb120_gazebo/launch/load_irb120.launch
+++ b/abb_irb120_gazebo/launch/load_irb120.launch
@@ -1,0 +1,3 @@
+<launch>
+  <param name="robot_description" command="$(find xacro)/xacro.py '$(find abb_irb120_gazebo)/urdf/irb120.xacro'" />
+</launch>
--- a/abb_irb120_gazebo/launch/load_irb120t.launch
+++ b/abb_irb120_gazebo/launch/load_irb120t.launch
@@ -1,0 +1,3 @@
+<launch>
+  <param name="robot_description" command="$(find xacro)/xacro.py '$(find abb_irb120_gazebo)/urdf/irb120t.xacro'" />
+</launch>
--- a/abb_irb120_gazebo/package.xml
+++ b/abb_irb120_gazebo/package.xml
@@ -1,0 +1,40 @@
+<package>
+  <name>abb_irb120_gazebo</name>
+  <version>0.0.1</version>
+  <description>
+    <p>
+      ROS-Industrial Gazebo support package for the ABB IRB 120 (and variants).
+    </p>
+    <p>
+      This package contains the configuration data and launch files required
+      to simulate the ABB IRB 120 manipulator in Gazebo. This includes the base
+      model and the 120T.
+    </p>
+    <p>
+      Before using any of the configuration files included in this package, be
+      sure to check they are correct for the particular robot model and
+      configuration you intend to use them with.
+    </p>
+  </description>
+  <author email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</author>
+  <author>Kevin Kelly (Trinity College Dublin)</author>
+  <maintainer email="culletom@tcd.ie">Mark Culleton (Trinity College Dublin)</maintainer>
+  <license>Apache2</license>
+
+  <url type="website">http://wiki.ros.org/abb</url>
+  <url type="bugtracker">https://github.com/ros-industrial/abb_experimental/issues</url>
+  <url type="repository">https://github.com/ros-industrial/abb_experimental</url>
+
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <run_depend>abb_irb120_support</run_depend>
+  <run_depend>gazebo</run_depend>
+  <run_depend>gazebo_ros</run_depend>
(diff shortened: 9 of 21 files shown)


gavanderhoorn:
@culletom: this is some nice work, thanks.

First question: any particular reason you went with effort controllers in this package?

Most industrial robots expose only position control to users, so the sim pkgs providing only those seems not too restrictive. Also: ~~it would probably mean less tweaking with pid gains (which will never completely match the 'real thing' anyway)~~ it avoids having to keep tweaking pid gains (as they are not needed). Finally, it would remove the need for the (fictional) effort limits in the urdfs.

culletom:
@gavanderhoorn That is a valid point, but from what I can see the default gazebo_ros_control plugin currently only supports effort controllers (looking at the [gazebo website](http://gazebosim.org/tutorials/?tut=ros_control#AddtransmissionelementstoaURDF)). If this is not the case though I will switch over to position control.

gavanderhoorn:
Well in ros-industrial/universal_robot#161, the controllers were switched to the position variants. I'm not too sure, but that would seem to indicate it can be done.

culletom:
I had a look at the link, but the suggested changes don't seem to work with hydro (not on my system anyway) so I am assuming they are specific to indigo. I tried using the PositionJointInterface hardware interface with the IRB 120, but apparently you still need to define PID controller values (as discussed halfway through [ros-simulation/gazebo_ros_pkgs#135](https://github.com/ros-simulation/gazebo_ros_pkgs/pull/135). This can be done in a separate config file, as suggested [here](http://answers.gazebosim.org/question/5468/gazebo_ros_pkgs-no-p-gain-specified-for-pid-in/), so I can do this if preferable.

culletom:
I've just updated the PR so that position controllers are used. As discussed in [ros-industrial/universal_robot#179](https://github.com/ros-industrial/universal_robot/pull/179), this causes error messages to be displayed during the gazebo startup, but these can be ignored.

gavanderhoorn:
I'll probably have a chance to test this PR sometime next week.

Personally I don't really mind the error messages, but it is probably a good idea to give users an advanced warning about them.

culletom:
Yea I agree - I added a comment about it in the readme file, but I could add an additional warning if required.

gavanderhoorn:
So I finally had time to play with/look at this PR (insert-emoticon-which-conveys-shame).

@culletom: I like it, and I believe we should do this (basic Gazebo compatible simulation pkgs) for all ros-i supported robot models. Setting up a `ros_control` compatible PositionController interface is really not so difficult, and it allows for some simple simulations to work out-of-the-box. Of course, without accurate info on the robot dynamics, the (use of such) simulations will be limited, but that is a different topic.

As to the contents of this PR, I propose the following:
1. move all Gazebo related files and content to the `irb120_gazebo` pkg and keep the current `irb120_support` package simulator-agnostic. I know we currently only 'aim' for Gazebo support, but I feel we shouldn't include any Gazebo-specific stuff in the main urdf/xacro in the support pkg. The `dynamics` and `inertial` tags will have to be added to the main urdf (Gazebo does not support adding those with a _gazebo reference_ or something similar), but those are properties of the manipulator anyway (sort-of), and not specific to Gazebo.
2. remove the `world` link and associated joint from the main Gazebo-xacro: it's needed if you want to spawn the model by itself in Gazebo, but if the Gazebo-xacro is to be included in another file, it hampers composibility (I've seen users ask/complain about that with the UR5/10 Gazebo pkgs fi). Perhaps we can add it to the `$model_gazebo/urdf/$model.xacro` file or something.
3. provide a separate `moveit_planning_execution_gazebo.launch` in the MoveIt Configuration pkgs:
   1. makes it immediately clear that it is setup for interaction with the corresponding Gazebo simulation of the manipulator
   2. removes the need for all the `if/unless` attributes and the `roslaunch` argument: arguments for launch files are unfortunately something many people have real difficulty with in my experience (discovering _and_ using them), so the fewer the better imo.
   3. a separate file would also allow us to prune the it a bit, as some nodes are not necessary when using Gazebo.
   4. finally, I can imagine similar files for V-Rep, etc in the future.

Things to figure out:
1. should the `gazebo_ros_control` plugin (and everything else) be namespaced (say with `mfg_model`)? I'd say yes, but I can't completely oversee what the advantages/implications would be compared to pushing nodes down into namespaces using `roslaunch`.
2. how to conveniently re-use any already existing urdf `material` and `color` definitions as Gazebo materials? According to the [sdf docs](http://sdformat.org/spec?ver=1.5&elem=material), it should be possible to add `diffuse` and / or `ambient` elements to links using a _gazebo reference_ in the urdf, but I've not been able to make that work (although I did only some quick testing). I'd rather avoid having to maintain two duplicates of essentially the same information (although the `mfg_resources` pkgs do make good places to put Gazebo material scripts).
3. if we do provide a `moveit_planning_execution_gazebo.launch` file, do we add a `run_depend` on the corresponding `mfg_model_gazebo` pkg to the MoveIt config's manifest? This would mean that installing the MoveIt cfg pulls in Gazebo et al as well, something which may not be desirable. Alternatively, we make it a convenience launch file which will only work correctly if the user has installed the required pkgs (unfortunately ROS has no concept of _suggested_ or _optional_ dependencies).

Note that I don't expect you to do the above: I'll submit a PR against your PR (`gazebo-devel` branch) some time this week.

Also, some questions / comments:
- why did you include the `irb120(t).world` files in the `_gazebo` package? They seem to be identical to the empty world file that is part of gazebo, apart from the world name. Can we not use that to spawn the model in?
- why did you set `respawn=false` on the `robot_state_publisher` node in `irb120_control.launch`?
- the way the current `irb120_support` pkg is setup (references to `gazebo_ros_control` plugin, etc), the pkg manifest seems to be missing `run_depend`s on `gazebo_ros` et al.

culletom:
@gavanderhoorn no need for the shame emoticon, I know you are busy so thanks for having a look! :smile: Its been a while since I have got the chance to work on this too, but the suggestions you have made make sense. I remember at the time trying to decide whether to keep all the gazebo files within the one folder or integrate them with the others, so I guess I made the wrong choice!

I have some time free this week so I will go ahead and update the PR. It needs to be tidied into a single commit anyways so I can do it all at once. I'll see if I can come up with anything for the 'things to figure out', but if not I will leave them for further discussion. The `irb120(t).world` files are not needed, I was using them at an earlier stage to add objects to world, so I will remove them. The other two questions / comments are mistakes on my part, so I will change them too.

gavanderhoorn:
@culletom: I've pushed some things I changed after looking at your PR to my personal fork. See the PR against yours. You don't need to accept it, but perhaps you can use it as inspiration.

Things I'm not happy with in my PR:
- not sure whether `moveit_planning_execution_gazebo.launch` should actually be part of `abb_irb120_gazebo` instead (and then drop the `_gazebo` suffix)
- now that base xacro macro has `inertial` elements, we get complaints from moveit/kdl_parser that the root link `base_link` has inertia configured
- perhaps loading of the IRB 120(T) Gazebo xacro should not be part of the `MODEL_gazebo.launch` (only load & init Gazebo in that file)
- should `robot_state_publisher` be started in `MODEL_control.launch`? The node is needed, but I'm not sure if it is the responsibility of something that initialises the `ros_control` controllers to start it.
- the `ros_control` ctrlr_mgrs don't shut down nicely when using convenience launch files now. Not sure what is going on, perhaps shutdown is not properly sequenced

gavanderhoorn:
Also:

> I remember at the time trying to decide whether to keep all the gazebo files within the one folder or integrate them with the others, so I guess I made the wrong choice!

Obviously, there is no _right_ or _wrong_ here. This is just what I think (based on experience) makes the most sense. Feel free to disagree (but with rationale / arguments).

culletom:
@gavanderhoorn ok thanks I will go through it now

culletom:
@gavanderhoorn thanks that it a good starting point. I agree with most changes, but I suggest keeping the `robot_state_publisher` node definition in the `MODEL_control.launch` for those who are wanting to control the robot within gazebo without moveit. Also I suggest not loading `irb120_gazebo.launch` and `irb120_control.launch` in the `moveit_planning_execution_gazebo.launch` file. I have been launching these in separate terminals, as the `ros_control` ctrlr_mgrs can then be shut down cleanly by terminating `irb120_control.launch` before `irb120_gazebo.launch`.

gavanderhoorn:
>  I suggest keeping the `robot_state_publisher` node definition in the `MODEL_control.launch` for those who are wanting to control the robot within gazebo without moveit.

I agree `robot_state_publisher` needs to be started, but as I wrote, I don't really like that something called `MODEL_control.launch` is also responsible for starting a seemingly unrelated node. Perhaps move it to the `MODEL_gazebo.launch` file or even create an additional file that acts as a coordinator of the two others?

> Also I suggest not loading `irb120_gazebo.launch` and `irb120_control.launch` in the `moveit_planning_execution_gazebo.launch` file. I have been launching these in separate terminals, as the `ros_control` ctrlr_mgrs can then be shut down cleanly by terminating `irb120_control.launch` before `irb120_gazebo.launch`.

Yes, I wasn't too happy with that either, but I wanted to include a 'one-click' solution for starting up everything (similar to `moveit_planning_execution.launch` itself).

As I wrote in the comments of `moveit_planning_execution_gazebo.launch`, I expect knowledgeable users to actually not use that file in more long-term setups (and the first thing then is probably to do what you suggest). Not sure what the best approach is though: having to start things in three separate terminals doesn't exactly seem too smooth ..

culletom:
I've rearranged the gazebo launch files there - its now a two stage process in order to launch everything (i.e. launch gazebo and then launch moveit) but I think this is ok. I moved the `robot_state_publisher` node definition to the `MODEL_gazebo.launch` file, which after checking matches a number of other gazebo packages. Alternatively we could follow the layout used in the [ur_gazebo](https://github.com/ros-industrial/universal_robot/tree/indigo-devel/ur_gazebo) package.

gavanderhoorn:
Just noticed you already mentioned the issue with the controller spawners hanging in your OP. Adding a `--shutdown-timeout 1` to their launch `args` makes things shutdown in an acceptable manner (no more `escalating to SIGTERM` etc).

> I've rearranged the gazebo launch files there - its now a two stage process in order to launch everything (i.e. launch gazebo and then launch moveit) but I think this is ok.

Yes, this is fine. In principle, I try to separate responsibilities as much as possible. Your current setup allows for someone to use the gazebo pkg stand-alone, as well as with the MoveIt configuration.

gavanderhoorn:
An additional comment: perhaps we should approach this differently. I'm thinking to treat the 'gazebo version' of a robot model as just that: another robot supported by ros-i. At the ROS API level, that robot should then implement all required interfaces, which would allow us to use it as 'just another ros-i robot'.

We'd need to see whether this is technically do-able with Gazebo (perhaps it'd require a custom plugin). Ideally, this would allow simulation of ros-i cells with Gazebo (or at least, robots).

For now I think just always remapping the Action Server topic exposed by the `gazebo_ros_control` plugin from `/whatever_namespace/follow_joint_trajectory` to `/joint_trajectory_action` (and not the other way around) would be a good first step. The other requirements are documented in the [ROS-Industrial Robot Driver Specification](http://wiki.ros.org/Industrial/Industrial_Robot_Driver_Spec).

gavanderhoorn:
> At the ROS API level, that robot should then implement all required interfaces, which would allow us to use it as 'just another ros-i robot'.

Playing with this a bit it would make the 'Gazebo version' of the `moveit_planning_execution.launch` files essentially identical to the normal ones, except for the part where the `planning_context.launch` gets included (we don't want it to load the robot xacro). O, and it doesn't need to load the `industrial_robot_simulator` nodes _or_ the real driver.

It would probably have been better if `moveit_planning_execution.launch` files did not `<include>` everything and would just load moveit (and perhaps rviz), but it's a bit too late for that now.

gavanderhoorn:
@culletom: have you been able to 'play around' with this some more? What do you think about the layout of the pkgs and the distribution of functionality?

culletom:
@gavanderhoorn apologies for the delay in response, I have been away for the last week. I haven't been able to do much with the package, but I did remap the action server as shown in the last commit. However, since we are using the Gazebo `ros_control` interface some parameters still exist with the `arm_controller` namespace. I'm not sure if this is a problem, but if so it may require the development of a custom Gazebo plugin as you previously suggested. 
Otherwise I think the layout of the packages works well - it seems to be a good fit with minimal interference to the original packages. One thing that you mentioned though is that the base link now has an inertial, which gives that warning during the moveit launch.

gavanderhoorn:
No problem. I was just wondering whether you'd run into something with the new layout. I think there are some minor things that could use some tweaking, and I believe the `transmission` markup should be changed slightly to be compatible with Indigo, but other than that this starts to look like a good way to structure Gazebo pkgs I believe.

culletom:
Yea I had looked at the `transmission` markup, but they differ between Hydro and Indigo so I decided to stick with the former as that was the version I was using at the time. But since Indigo is the LTS version, I'll change this now.

gavanderhoorn:
I think it makes the most sense: if they ever get released, it will be for `ros_control` / `gazebo_plugins` versions >= Indigo, so that will require the new markup.

Users still on Hydro should fork and adapt.

culletom:
Yea I agree, I have just brought in those changes there now.

gavanderhoorn:
May I suggest to merge this as it is now (in an `indigo-devel` branch)? As this is an experimental repository, we can easily iterate on it, and it seems this is a nice piece of functionality.

@culletom: the only thing I'd change is to not use the `controller_manager/spawner` in `irb120(t)_control.launch`, but to use `controller_manager/controller_manager spawn <ctrlr_name>`. That would remove the need for the `--shutdown-timeout` parameter, and is iiuc, cleaner for this use case (see ros-controls/ros_control#188).

In any case, if you could squash all commits into a single one @culletom, that would be nice.

gavanderhoorn:
O and perhaps move the Gazebo related info that is currently in the main repository `README.md` to a file in `abb_irb120_gazebo` itself (and then also [install that file](https://github.com/ros-industrial/fanuc/blob/67e17e8ef7d6a32e842a3ae274ad7819e3b3d5c1/fanuc_m10ia_support/CMakeLists.txt#L27)).

culletom:
Thats sounds good to me, I'll make those suggestions changes and squash
them to the one commit. Should be ready later today.

On 6 September 2015 at 15:05, G.A. vd. Hoorn notifications@github.com
wrote:

> May I suggest to merge this as it is now (in an indigo-devel branch)? As
> this is an experimental repository, we can easily iterate on it, and it
> seems this is a nice piece of functionality.
> 
> @culletom https://github.com/culletom: the only thing I'd change is to
> not use the controller_manager/spawner in irb120(t)_control.launch, but
> to use controller_manager/controller_manager spawn <ctrlr_name>. That
> would remove the need for the --shutdown-timeout parameter, and is iiuc,
> cleaner for this use case (see ros-controls/ros_control#188
> https://github.com/ros-controls/ros_control/pull/188).
> 
> In any case, if you could squash all commits into a single one @culletom
> https://github.com/culletom, that would be nice.
> 
> —
> https://github.com/ros-industrial/abb_experimental/pull/18#issuecomment-138087946
> .

## 

_Mark Culleton_

PhD Candidate
Robotics Research Group
Department of Mechanical & Manufacturing Engineering
Trinity College Dublin
Funding Body: Irish Research Council
E: culletom@tcd.ie
P: (+353) 86 1772095

culletom:
@gavanderhoorn I've updated the files as suggested and squashed the commits, so hopefully all is in order now.

gavanderhoorn:
I think something went wrong: there are now two commits:
- ec28127 containing your Gazebo changes, but also a lot of other stuff, and
- a464abb containing something related to `abb_resources`

culletom:
@gavanderhoorn Yes turned out it went very wrong! Ended up having to reset the branch and recommit, but it looks in order now

culletom:
@gavanderhoorn does this need to be +1'ed by someone else, or can it be merged now as it?

Levi-Armstrong:
It can be merged.

gavanderhoorn:
@Levi-Armstrong wrote:

> It can be merged.

Unfortunately, I believe we should've merged this into an `indigo-devel` branch, as the `abb_irb120_gazebo` package is really only compatible with `ros_control` versions >=Indigo.

@Levi-Armstrong: I suggest creating an `indigo-devel` branch and then removing `abb_irb120_gazebo` from `groovy-devel`.

Levi-Armstrong:
@gavanderhoorn: I will get the new branch created and remove this from groovy-devel

gavanderhoorn:
Great. Thanks.

@culletom: forgot to say: thanks for this contribution, and thanks for iterating on it with me while I was being difficult. Much appreciated!

PR #21, opened 2015-03-01
Fix #19: Updated abb_irb4400_support package

### LINKED ISSUE #19: abb_irb4400_support package needs updating
Urdf files need to be updated and formatted to new standard. This includes validating limit values and joint transforms.

culletom:
Corrected joint limits and velocities, and reformatted the urdf files. Also added additional files so that it matches other support packages.


culletom:
Just wanted to confirm that the joint transforms need to be modified before I went ahead and changed them, particularly between joint 1 and joint 2.

Levi-Armstrong:
They are acceptable, don't go through the trouble.

Levi-Armstrong:
@culletom, There is an existing error with the urdf that affects the kinematics, joint_4 x-axis translation should be 1.38 and joint_6-tool0 x-axis tranlation should be 0.14. This would also require a correction to the meshes.

abb_irb4400_support/launch/test_irb4400l30.launch
```diff
@@ -1,7 +1,7 @@
 <launch>
-	<include file="$(find abb_irb4400_support)/launch/load_irb4400l30.launch" />
-	<param name="use_gui" value="true" />
-	<node name="joint_state_publisher" pkg="joint_state_publisher" type="joint_state_publisher" />
-	<node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher" />
-	<node name="rviz" pkg="rviz" type="rviz" args="-d $(find industrial_robot_client)/config/robot_state_visualize.rviz" required="true" />
+  <include file="$(find abb_irb4400_support)/launch/load_irb4400l30.launch" />
+  <param name="use_gui" value="true" />
```
gavanderhoorn: Make this a private param of the `joint_state_publisher` node.

abb_irb4400_support/package.xml
```diff
@@ -3,18 +3,17 @@
   <version>0.1.0</version>
   <description>
     <p>
-      ROS Industrial support for the ABB IRB4400 (and variants).
+      ROS Industrial support for the ABB IRB4400/L30.
```
gavanderhoorn: Please leave this "(and variants)". It is sort-of standardised across pkgs, and we won't have to change this when we do add more variants of the 4400 to the pkg.

abb_irb4400_support/package.xml
```diff
@@ -3,18 +3,17 @@
...
     </p>
     <p>
       This package contains configuration data, 3D models and launch files 
-      for ABB IRB4400 manipulators. This currently includes the L30.
+      for ABB IRB4400/L30 manipulator.
```
gavanderhoorn: See earlier comment.

abb_irb4400_support/package.xml
```diff
@@ -3,18 +3,17 @@
...
     <p>
-      Joint limits and max joint velocities are based on the information in 
-      the ABB data sheets.  All URDFs / XACROs are based on the 
-      default motion and joint velocity limits, unless noted otherwise (ie: 
-      no support for high speed joints, extended / limited motion ranges or 
-      other options).
+      Joint limits and max joint velocities are based on the information in the <a href="http://www05.abb.com/global/scot/scot352.nsf/veritydisplay/c90b98aaa057bd6ec12576cb00528ef6/$file/Product%20specification%204400%20M99%20BWOS3.2.pdf">ABB 
+      IRB 4400 product specification document</a> (Article No: 3HAC 8770-1).  All URDFs / XACROs are based on the 
+      default motion and joint velocity limits, unless noted otherwise (ie: no support for high speed joints, 
+      extended / limited motion ranges or other options).
```
gavanderhoorn, line 18: It's almost impossible, but please try to keep this text within 80 columns. The URL is obviously too long, but it should be possible for the rest of the text.

Also, please lowercase "URDFs / XACROs".

abb_irb4400_support/launch/robot_state_visualize_irb4400l30.launch
```diff
@@ -0,0 +1,27 @@
...
+
+  Defaults provided for IRB 4400/L30:
+   - J23_coupled = true
+
+  Usage:
+    robot_state_visualize_irb4400l30.launch robot_ip:=<value>
+-->
+<launch>
+  <arg name="robot_ip" />
+  <arg name="J23_coupled" default="true" />
```
gavanderhoorn: I'm sure you have, but just to be sure: you have verified this with robot studio or a real robot?

gavanderhoorn:
Changes look good, but some minor things need to be addressed.

Please also check the comment by @Levi-Armstrong about link lengths for `link_4` and the `tool0` transform.

Levi-Armstrong:
I have verified with RobotStudio that J23 is coupled.

Also, I would prefer the naming convention to be model_payload_reach. This will cover all possible variants for the abb robots, so for this variant it would be irb4400l_30_243. @shaun-edwards & @gavanderhoorn what are your thoughts?

gavanderhoorn:
The naming should definitely include all information needed to uniquely identify a manipulator variant. For the ABB that appears to be model nr, payload and reach.

Should the `L` be grouped with the model nr though, or with the `30`? Looking at [abb-irb4400-data](http://new.abb.com/products/robotics/industrial-robots/irb-4400/irb-4400-data) it looks like it should be grouped with the `30` ('IRB 4400-L10', on that page). It would then become `abb_irb4400_l30_243`?

gavanderhoorn:
For other models ([irb 1600](http://new.abb.com/products/robotics/industrial-robots/irb-1600/irb-1600-data) fi), we would get: `abb_irb1600_10_120` (for the '1600 - 10 / 1.2') and `abb_irb1600_6_145` (for the '1600 - 6 / 1.45').

The 'rule' would then be to append the payload in kgs, and the reach in meters, separated by underscores. And always use at least two decimal places after the dot for the reach (so `120`, instead of `12`).

Levi-Armstrong:
In RobotStudio they group the nr with the model, I think this would be better so not to confuse people. The confusion being they may think the L has some significant related to the payload when it actual stands for long reach.

Levi-Armstrong:
I agree.

> The 'rule' would then be to append the payload in kgs, and the reach in meters, separated by underscores. And always use at least two decimal places after the dot for the reach (so 120, instead of 12).

gavanderhoorn:
> In RobotStudio they group the nr with the model, I think this would be better so not to confuse people. The confusion being they may think the L has some significant related to the payload when it actual stands for long reach.

Ok. Agreed. I was just going with what was listed on the ABB site.

Levi-Armstrong:
Yea, that is what I original found too.

On Fri, Mar 6, 2015 at 2:59 AM, G.A. vd. Hoorn notifications@github.com
wrote:

> In RobotStudio they group the nr with the model, I think this would be
> better so not to confuse people. The confusion being they may think the L
> has some significant related to the payload when it actual stands for long
> reach.
> 
> Ok. Agreed. I was just going with what was listed on the ABB site.
> 
> —
> https://github.com/ros-industrial/abb_experimental/pull/21#issuecomment-77526564
> .

## 

Levi Armstrong

culletom:
Ok I'll update the PR to include this naming convention, and the other comments mentioned. Should file names also be updated, or is it just within the urdf / xacros?

gavanderhoorn:
Everything needs to be consistent.

As this is a package in an experimental repository, that is allowed without consideration for any bw-compatibility.

gavanderhoorn:
We should clarify that for support pkg naming, no variant names should be used. With the grouping of the `L` with the model nr, contributors could get that idea, so we might want to make sure we document this somewhere.

Levi-Armstrong:
@gavanderhoorn & @shaun-edwards, would [this](http://wiki.ros.org/Industrial/Tutorials/WorkingWithRosIndustrialRobotSupportPackages#Naming) be the best place for documenting the naming convention for the abb robots?

gavanderhoorn:
I'd like to keep that page about working with the pkgs. We could update [File and directory layout for robot support repositories](http://wiki.ros.org/Industrial/Tutorials/SuggestedPackageLayoutNewRepositories) (or I should finish [the rep version](https://github.com/gavanderhoorn/rep/blob/repository_layout/rep-xxxx.rst) some time and add it there, as an appendix or something).

gavanderhoorn:
Possible text for ABB specific naming rules for the linked wiki page/rep:

---

# Appendix A: Vendor specific naming

## ABB

Variant names for ABB manipulators may include a _modifier_ (indicating support for a particular feature, or a physical characteristic of the variant), a _payload specification_ (one or more digits, indicating maximum supported payload in kg) and finally a specification of the _reach_ of the arm (in meters). Examples: _IRB 1600 - 10 / 1.2_ and _IRB 4400L - 10_.

All ABB support packages shall be named using the following template:

```
  abb_irbMODEL_support
```

Note that this template omits the modifier, payload and reach components included in the actual product name.

All artefacts within ABB support, MoveIt configuration and plugin packages shall be named using the following template:

```
  abb_irbMODEL[MODIFIER]_PAYLOAD_REACH
```

Where MODEL is a numeric string, MODIFIER is a string, PAYLOAD is a numeric string (max payload in kg) and REACH is a numeric string (reach converted to centimeters, avoiding fractional numbers).

If a particular model does not have variants based on reach, the REACH component should be omitted from the ROS-Industrial name as well.

Examples: `abb_irb1600_10_120` and `abb_irb4400l_10`.

---

@Levi-Armstrong: does that make sense?

culletom:
@gavanderhoorn I've renamed all files to follow your outlined convention, but I will wait for confirmation from @Levi-Armstrong before committing. 

One thing I have yet to fix is the error with the x-axis translations, as I can't find the CAD model for the L30 on the ABB website and so I'm not sure how to modify the meshes. Should I translate the mesh origins using something like blender, or is there an easier approach?

Levi-Armstrong:
@gavanderhoorn, in reference to the vendor specific naming, I would suggest that the naming always include the reach. In the case of the irb 4400l there is two payloads variants 10kg and 30kg and they actual have different reaches; the 10kg is 2.55m and the 30kg is 2.43m. The suggest naming for the variants would be abb_irb4400l_10_255 and 4400l_30_243. Do you agree?

Levi-Armstrong:
@culletom; I would use meshlab to translate the mesh; the command is under Filters>Normals, Curvatures and Orientation>Transform: Move, Translate, Center. Make sure the freeze matrix is checked, if not when you save the mesh the changes will not be applied.

gavanderhoorn:
@Levi-Armstrong wrote:

> @gavanderhoorn, in reference to the vendor specific naming, I would suggest that the naming always include the reach. In the case of the irb 4400l there is two payloads variants 10kg and 30kg and they actual have different reaches; the 10kg is 2.55m and the 30kg is 2.43m. The suggest naming for the variants would be abb_irb4400l_10_255 and 4400l_30_243. Do you agree?

Hm, not sure. I've always approached these kind of issues such that the minimum amount of information should be used to discriminate between two variants. In the case you describe, the payload already uniquely identifies the variant, so including reach would seem to be unnecessary. Also, the ABB model/variant name doesn't include the reach for those models, which would make the ros-i name different.

But perhaps we should split this discussion of into a separate thread/issue, as it's getting a bit off-topic for this PR.

gavanderhoorn:
Let's continue on ros-industrial/abb#75 with the naming discussion.

culletom:
@Levi-Armstrong thanks for the advice, made the translations much easier. 

The last commit brings in all the suggested changes. I also renamed the files, but depending on the outcome of the naming discussion these might need to be renamed again. However if this ends up being the case it can be done later in a sperate PR, as other packages will have to be renamed too.

gavanderhoorn:
@culletom: yeah, I'm sorry about the naming 'mess'. Thanks for sticking with us though, really appreciated.

---

As to your last commit: as shown [here](http://wiki.ros.org/Industrial/Tutorials/SuggestedPackageLayoutNewRepositories#Support), the `meshes/VARIANT/` dir should not include the mfg prefix.

This is probably not too apparent from the linked text, but the rationale is that only entities that are referenced externally / included in other files _without the package name included in the URI_ need the mfg prefix, otherwise it is redundant.

As an example: a xacro macro file can only be included in another xacro by referring to it like `package://mfg_pkg/path/to/the_macro.xacro` or `$(find mfg_pkg)/path/to/the_macro.xacro`. This already includes the `mfg_pkg` in the URI, so adding it again to `the_macro.xacro` is unneeded (as there is no chance of clashes).

_But_, the actual robot macro that is defined by the macro _should_ get the prefix, as it will be invoked as `<xacro:robot_macro />`, which does not include the mfg prefix, so if there are two mfgs that have a `robot` model/variant, we can get clashes, hence the requirement for the prefix.

I hope it now makes sense that the `meshes/VARIANT` dir does not need the prefix. It can simply be named `irb4400l_30`. If in the future more variants are added, and those can share meshes with the L-30, we should probably create a `meshes/irb4400`, move common meshes there and reference them from the variant specific urdfs/xacros.

---

I guess I should spend some time updating the documentation on support pkg layout.

We do also have a [support pkg generator](https://github.com/ros-industrial/industrial_experimental/tree/hydro-devel/industrial_robot_pkg_gen), although it has not been widely 'advertised'.

culletom:
@gavanderhoorn no problem, naming has come up in a number of different PRs, so its important to settle once and for all! 

Yes what you have suggested makes perfect sense - I've updated the files to reflect this.

Levi-Armstrong:
@culletom, The package.xml file is missing the <build_depend>roslaunch</build_depend>, other than that it looks pretty good. Thank you for contributing.

abb_irb4400_support/CMakeLists.txt
```diff
@@ -6,7 +6,12 @@ find_package(catkin REQUIRED)
 
 catkin_package()
 
+if (CATKIN_ENABLE_TESTING)
+  find_package(roslaunch REQUIRED)
+  roslaunch_add_file_check(tests/roslaunch_test.xml)
```
gavanderhoorn: Please add a `DEPENDENCIES` argument to the `roslaunch_add_file_check(..)` invocation. See [here](https://github.com/ros-industrial/fanuc/blob/5e2cf28d4b65af0724566a132c0f56aff6f737d2/fanuc_m10ia_support/CMakeLists.txt#L11-L18) for an example.

abb_irb4400_support/package.xml
```diff
@@ -32,9 +33,11 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
-  <!-- Typical run depends, not yet populated
-  <run_depend>abb_assets</run_depend>
+  <build_depend>roslaunch</build_depend>
```
gavanderhoorn: If you add the `DEPENDENCIES` arg (see earlier comment), we should state the `roslaunch` version here, just to be safe. See [here](https://github.com/ros-industrial/fanuc/blob/5e2cf28d4b65af0724566a132c0f56aff6f737d2/fanuc_m10ia_support/package.xml#L36) for an example.

gavanderhoorn:
@culletom: pending acceptance of the new naming guidelines for ABB pkgs, I'd say this is pretty complete now. If you address the two minor points related to the `roslaunch` dependency and use, I think we can merge this as it is right now.

If you could squash all fixups into the first commit that would be even nicer (if you want you can extend the commit msg to keep a record of what was changed).

culletom:
@gavanderhoorn I've just included the `roslaunch` dependencies and consolidated the PR into one commit. With regards to the `roslaunch` dependencies, these currently aren't included in other abb support packages, so perhaps this needs to be done?

gavanderhoorn:
We'll do the other pkgs when they receive other updates. I'll leave it to @Levi-Armstrong to determine whether he wants to _roslaunch test_ all support pkgs before releasing into Indigo or not.

Levi-Armstrong:
@culletom: Which support packages are missing the roslaunch dependencies?

Levi-Armstrong:
+1

culletom:
@Levi-Armstrong I was referring to the points brought up by @gavanderhoorn (see the last two 'outdated' line notes above), which highlight the need to include a `DEPENDENCIES` argument in the roslaunch_add_file_check(..) invocation in the CMakeList.txt file. This is currently not done in any of the other abb support packages.

PR #23, opened 2015-03-15
Update materials

Levi-Armstrong:


gavanderhoorn:
Visual review is ok. Nice one fixing this @Levi-Armstrong: good to have some consistency with these things.

Levi-Armstrong:
@gavanderhoorn, So it is good to merge?

gavanderhoorn:
Yes, I think it is. As I said, I've only done a visual review (so haven't tested by launching any `test_...launch` files).

+1.

PR #25, opened 2015-07-25
Add ABB IRB6600 225kg 2.55m

Levi-Armstrong:


abb_irb6600_support/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+
+if (CATKIN_ENABLE_TESTING)
+  find_package(roslaunch REQUIRED)
+  roslaunch_add_file_check(tests/roslaunch_test.xml)
+endif()
+
+foreach(dir config launch meshes urdf)
+   install(DIRECTORY ${dir}/
+      DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/${dir})
+endforeach()
```
gavanderhoorn: 3 vs 2 spaces.

gavanderhoorn: Also: this can be simplified:

``` cmake
install(DIRECTORY config launch meshes urdf
  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
```

No need for the for-loop anymore.

abb_irb6600_support/launch/robot_interface_download_irb6600_225_255.launch
```diff
@@ -0,0 +1,20 @@
+<!--
+  Manipulator specific version of abb_driver's 'robot_interface.launch'.
+
+  Defaults provided for IRB 6600:
+   - J23_coupled = false
```
gavanderhoorn, line 5: This (coupling) is probably true for all `6600` variants, but perhaps be specific here as well (so: "Defaults provided for IRB 6600 - 225/2.55")?

abb_irb6600_support/launch/robot_state_visualize_irb6600_225_255.launch
```diff
@@ -0,0 +1,27 @@
+<!--
+  Manipulator specific version of the state visualizer.
+
+  Defaults provided for IRB 6600:
+   - J23_coupled = false
```
gavanderhoorn, line 5: See earlier comment

abb_irb6600_support/package.xml
```diff
@@ -0,0 +1,41 @@
...
+      Before using any of the configuration files and / or meshes included
+      in this package, be sure to check they are correct for the particular
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author email="levi.armstrong@swri.org">Levi Armstrong</author>
+  <maintainer email="levi.armstrong@swri.org">Levi Armstrong</maintainer>
+  <license>Apache2</license>
+
+  <url type="website">http://wiki.ros.org/abb</url>
```
gavanderhoorn, line 29: Should we create a `abb_experimental` page on the wiki, similar to `fanuc_experimental` etc? With all the development going on in this repository, might be a good idea.

abb_irb6600_support/urdf/irb6600_225_255_macro.xacro
```diff
@@ -0,0 +1,348 @@
...
+      <origin xyz="0 0 0" rpy="0 0 0" />
+      <parent link="${prefix}link_1" />
+      <child link="${prefix}link_1_mtr" />
+    </joint>
+    <joint name="${prefix}joint_1_mtr2" type="fixed">
+      <origin xyz="0 0 0" rpy="0 0 0" />
+      <parent link="${prefix}link_1" />
+      <child link="${prefix}link_1_mtr2" />
+    </joint>
+    <joint name="${prefix}joint_1_wt" type="fixed">
```
gavanderhoorn, line 267: Cosmetic only: parallel bar structures can be modelled using [mimic joints](http://wiki.ros.org/urdf/XML/joint). Might be nice to do that here as well. I haven't investigated how that interacts with MoveIt though, so that might be something to test.

See the [M-900iA/260L](https://github.com/ros-industrial/fanuc_experimental/blob/3ef27dbf1402ee7e765479a6a48b1de530cec06b/fanuc_m900ia_support/urdf/m900ia260l_macro.xacro#L195-L208) for an example.

gavanderhoorn:
Some minor comments, but very nice PR @Levi-Armstrong.

The major issue here of course is that we already have an `abb_irb6600_support` pkg (in the `abb` repository), which should really be called `abb_irb6640_support` (unfortunate mistake by me, in ros-industrial/abb#43).

I'm not sure how to deal with this: ideally we'd rename `abb_irb6600_support` to `abb_irb6640_support`, but that pkg has been released already (and into Indigo, so LTS). 

All 6640 related files do include the `40` in their name, so it would be (theoretically) possible to merge the pkg proposed in this PR and the existing one without issues. The `irb6640` model would be in a wrong pkg, but that we can fix in a new pkg which is essentially a rename of the current `abb_irb6600_support` pkg (which would contain copies of the files).

Thoughts?

gavanderhoorn:
Introducing a new pkg for the IRB6640 would also allow us to be more specific about which variant it actually is ([manual](http://www.abbrobots.co.uk/en/3HAC026876-en.pdf)).

abb_irb6600_support/launch/robot_interface_download_irb6600_225_255.launch
Levi-Armstrong, line 5: This was original true but after testing this model on the actual robot it was found not to be the case.

Levi-Armstrong:
I think the approach you laid out would be best, to create a new package abb_irb6640_support (copy of the current abb_irb6600_support in the abb repository) and correct the naming to follow the new naming convention. Then leave the 6640 in the abb_irb6600_support and deprecate these file to be removed in Jade.

Now, for this to work does this PR need to be submitted to the abb repository so there is not duplicate packages with the same name?

abb_irb6600_support/package.xml
Levi-Armstrong, line 29: Are you referring to something like [this](http://wiki.ros.org/fanuc_m900ia_support) for each package or am I looking in the wrong place?

abb_irb6600_support/launch/robot_interface_download_irb6600_225_255.launch
gavanderhoorn, line 5: I meant that it might be good to be specific about which variant this file is providing defaults for (so instead of just `IRB 6600`, `Defaults provided for IRB 6600 - 225/2.55`).

abb_irb6600_support/package.xml
gavanderhoorn, line 29: Sort-of: I meant: should we create a page for the `abb_experimental` repository? Whether we create pages for the various pkgs is then up to you/us. Just seems more correct to link user to `wiki.ros.org/abb_experimental`, instead of the main page, where they won't find these pkgs.

We could then also document how to 'install' the experimental pkgs on that page.

gavanderhoorn:
> Now, for this to work does this PR need to be submitted to the abb repository so there is not duplicate packages with the same name?

Yeah, that would be necessary, unfortunately.

Would perhaps be better to first get the IRB 6640 sorted out (e.g.: create 6640 support pkg, etc), and _then_ re-open this PR against `abb`, merging in the actual 6600 this is introducing.

Levi-Armstrong:
@gavanderhoorn I have created a [pull request](https://github.com/ros-industrial/abb/pull/89) where I created a copy of the abb_irb6600_support and renamed everything according to the new naming convention.

Levi-Armstrong:
This PR has been open against ros-industrial/abb repository.

PR #39, opened 2017-01-18
Remove abb_resources package to be transferred to abb repo

Levi-Armstrong:
@gavanderhoorn, Can you please review when available?


gavanderhoorn (approved):

gavanderhoorn:
Looks ok to me.

Travis fails because obviously this pkg is now nowhere to be found.

After this is merged, could you update ros-industrial/abb#118 to reference f3816a1? Thay way we keep a consistent history.

gavanderhoorn:
Note that we'll have to update the Travis configuration in this repository to pull `abb` from sources instead of relying on released pkgs, at least until `abb_resources` has been released from the `abb` repository.

See [staubli_experimental/.travis.yml](https://github.com/ros-industrial/staubli_experimental/blob/87075673c61acca38a80616af5c9c513868a930a/.travis.yml) for how that can be done (in combination with [.travis.rosinstall](https://github.com/ros-industrial/staubli_experimental/blob/3c39640efe9b7f7f2794d00270d046c271afd504/.travis.rosinstall)).

PR #41, opened 2017-03-10
kinetic-devel release

AustinDeric:
This PR is intended to be merged after the ros-industrial/abb kinetic-devel branch.  This way we can reference the correct repo in the abb_experimental.rosinstall file.


.travis.rosinstall
```diff
@@ -1,3 +0,0 @@
-- git:
-    uri: https://github.com/ros-industrial/abb.git
-    local-name: ros-industrial/abb
```
gavanderhoorn: Is this no longer needed? Travis didn't work without it earlier, as `abb_resources` hasn't seen a release from `ros-industrial/abb` yet.

README.md
```diff
@@ -1,10 +1,21 @@
...
 
 [ROS-Industrial]: http://www.ros.org/wiki/Industrial
 [ROS wiki]: http://ros.org/wiki/abb_experimental
+
+## Build
+This build methods uses:
+ - [wstool](http://wiki.ros.org/wstool) for rosinstall.
+ - [catkin tools](https://catkin-tools.readthedocs.io/en/latest/) for building.
+```
+mkdir -p abb_experimentnal_ws/src
```
gavanderhoorn: Same typo as in ros-industrial/abb#126.

abb_experimental.rosinstall
```diff
@@ -0,0 +1,5 @@
+#Temporary until PR accepted by https://github.com/ros-industrial/
+- git: {local-name: abb, uri: 'https://github.com/AustinDeric/abb.git', version: kinetic-devel}
```
gavanderhoorn: Should users really clone your fork of `abb`?

abb_experimental.rosinstall
```diff
@@ -0,0 +1,5 @@
...
+
+- git: {local-name: industrial_core, uri: 'https://github.com/ros-industrial/industrial_core.git', version: kinetic-devel}
```
gavanderhoorn: Same comment as in ros-industrial/abb#126: `industrial_core` has a Kinetic release, so if possible could we use that?

abb_experimental/package.xml
```diff
@@ -8,9 +9,13 @@
   <author email="sedwards@swri.org">Shaun Edwards</author>
 
   <buildtool_depend>catkin</buildtool_depend>
-  <run_depend>abb_irb4400_support</run_depend>
+
+  <exec_depend>abb_resources</exec_depend>
+  <exec_depend>abb_driver</exec_depend>
+  <exec_depend>abb_irb4400_support</exec_depend>
```
gavanderhoorn: Did you have a particular reason to include only the `abb_irb4400_support` package here? And not the ones for the IRB 120?

Also: `abb_resources` and `abb_driver` are packages in another repository. Did you add those intentionally?

abb_irb120_support/CMakeLists.txt
```diff
@@ -12,8 +13,7 @@ if (CATKIN_ENABLE_TESTING)
     DEPENDENCIES
       # not nice, but industrial_robot_client doesn't prefix its targets
       joint_trajectory_action
-      abb_driver_robot_state
-      abb_driver_motion_download_interface)
+      abb_driver_robot_state)
```
gavanderhoorn: Is the `abb_driver_motion_download_interface` not needed anymore?

In any case: all `DEPENDENCIES` should probably be removed, it doesn't work (correctly) when doing isolated builds.

gavanderhoorn (changes requested):
Nice work.

One additional comment: none of the packages in this repository host any compilable artefacts, so is adding the `c++11` flags everywhere necessary?

README.md
AustinDeric: this one isn't a typo.  With that said, i can make it abb_ws/src, let me know which you prefer.

.travis.rosinstall
AustinDeric: The concept is needed, but I just put it straight into the .travis.yml. This essentially uses the rosinstall method to download the abb repo.
```
global:
    - ROS_DISTRO=kinetic
    - UPSTREAM_WORKSPACE=file
    - ROSINSTALL_FILENAME=abb_experimental.rosinstall
```

abb_experimental.rosinstall
AustinDeric: Only until we get the ros-industrial/abb kinetic-devel branch up and running.  My intent was for this to be merged after that so we can make the change in this file.  In hid-sight i should have said this in my description, i will fix.  

The reason i did it like this is so that we can test it in travis.

abb_experimental.rosinstall
AustinDeric: I can update the readme for the rosdep method if you want.  We will still have to use the rosinstall method to get the abb package.  So in effect we will have to do rosinstall + rosdep rather than just rosinstall. (i hope that made sense).

abb_irb120_support/CMakeLists.txt
AustinDeric: @Levi-Armstrong is abb_driver_motion_download_interface needed anymore?  I couldn't find it anywhere.

README.md
gavanderhoorn: Shouldn't it be `abb_experimental_ws` instead of `abb_experimentNal_ws`? The typo is repeated in all lines below as well.

abb_experimental.rosinstall
gavanderhoorn: <del>Personally I'm really not comfortable telling users in the official readme of `abb_experimental` to build a fork of `abb`. Would it be really a problem if we wait with this PR until your PR against `ros-industrial/abb` is merged?</del>

gavanderhoorn: <del>@Levi-Armstrong, what do you think?</del>

O wait, I misunderstood your comment. You intend to change that line back to `ros-industrial/abb` as soon as ros-industrial/abb#126 is merged.

abb_experimental.rosinstall
gavanderhoorn: >  So in effect we will have to do rosinstall + rosdep rather than just rosinstall

Doing a `rosdep install ..` after a `wstool merge ..` (which I think you meant) is regular development practice I think?

abb_irb120_support/CMakeLists.txt
gavanderhoorn: `abb_driver_motion_download_interface` is the name of the target in `abb/abb_driver/CMakeLists.txt` that builds the `motion_download_interface` binary.

But as I wrote, the whole `DEPENDENCIES` arg list should be removed as it doesn't work as intended.

abb_experimental.rosinstall
gavanderhoorn: @Levi-Armstrong what is your opinion on having `industrial_core` in the `.rosinstall` file versus just using the Kinetic release?

.travis.rosinstall
gavanderhoorn: Ok. That would work, as `abb` is now a source repo in the `.rosinstall` file. Reason a separate `.travis.rosinstall` was added is because it is/was the recommended way to get additional source dependencies in an `industrial_ci` Travis session while making it explicit that these are not normally needed to be cloned as sources in the workspace.

README.md
```diff
@@ -1,10 +1,21 @@
...
+cd abb_experimentnal_ws/src
+git clone -b kinetic-devel https://github.com/ros-industrial/abb_experimental.git
+wstool init .
+wstool merge abb_experimentnal/abb_experimentnal.rosinstall
+wstool update
```
gavanderhoorn: Same possibility for simplification as in ros-industrial/abb#126.

abb_experimental.rosinstall
Levi-Armstrong: >@Levi-Armstrong what is your opinion on having industrial_core in the .rosinstall file versus just using the Kinetic release?

@gavanderhoorn I think it would be best to default to using the release version.

AustinDeric:
changes based on the approval of https://github.com/ros-industrial/abb.  Needs minor adjustments in the rosinstall file when the kinetic-devel branch is corrected.  This PR is to show the current state of the branch.

PR #50, opened 2017-04-27
Intial commit for ABB irb1600 6kg 1.2m support package

AustinDeric:
based on #12 and work from  @Samsagax and @asct.

Here is a list of fixes that this PR required (thank you @gavanderhoorn):
 - [x] meshes have uppercase file extensions (keep everything lower case)
 - [x]  multiple references to IRB 2400 (launch files)
 - [x]  variant suffix (_12) seems not specific enough. Assuming 12 is actually '1.2', is this the 6 / 1.2 or 10 / 1.2 (from here)?
 - [x]  variant suffix should be used throughout (file names, content, directory names)
 - [x]  if meshes are usable for all models, meshes subdir should not include variant suffix (or at least split out the common ones)
 - [ ]  what is the 'base model' of the IRB 1600 (pkg manifest)?
 - [x]  which data sheets were used, specifically (url) (pkg manifest)?
 - [x]  the package is in abb_experimental, not in abb (pkg manifest)
 - [x]  urdf and xacro seem out-of-sync: orientation of tool0 is 0 1.57079 0 in urdf, but 0 .57079 0 in xacro?
 - [x] indentation in xacros should be two spaces per level
 - [x]  xml elements in xacro should be as much as possible formatted as single lines (origin elements fi), unless they have child elements (easily solved with something like xmlprettyprint.com
 - [x]  calibration and dynamics elements should be removed (or updated, 0 seems like default values)

Does anyone know if there are any differences between the 10kg and 6kg versions of the 1.2m version?  IF there are no differences, how do we handle that with the naming conventions: http://wiki.ros.org/Industrial/Tutorials/SuggestedPackageLayoutNewRepositories#General


gavanderhoorn:
@AustinDeric wrote:
> Does anyone know if there are any differences between the 10kg and 6kg versions of the 1.2m version? IF there are no differences, how do we handle that with the naming conventions

The obvious difference is the maximum handling capacity, and all the changes in dynamics that come with that. However, none of the motion planners in ROS currently take anything like that into account, allowing us to 'optimise' our support packages by just using the 6kg variant wherever the 10kg variant is needed.

This is not explicitly documented somewhere as of yet I believe. It is mentioned in the last paragraph of [Naming conventions for ABB robots in ROS-Industrial - Alternatives - Ignoring payload](https://github.com/ros-industrial/rep/blob/master/rep-I0007.rst#ignoring-payload). I also discussed this with @Levi-Armstrong in https://github.com/ros-industrial/abb/issues/75#issuecomment-276061154 and follow-up comments.

abb_irb1600_support/package.xml
```diff
@@ -0,0 +1,42 @@
...
+  <description>
+    <p>
+      ROS-Industrial support for the ABB IRB 1600 (and variants).
+    </p>
+    <p>
+      This package contains configuration data, 3D models and launch files
+      for ABB IRB 1600 manipulators. This currently includes the base model.
+    </p>
+    <p>
+      Joint limits and max joint velocities are based on the information in
```
Levi-Armstrong: Can you add a link to the product specification for the 1600?

https://library.e.abb.com/public/818d1922d857a640c1257cc500449f38/Product%20specifications_IRB1600_3HAC023604-001_revM_en.pdf

Levi-Armstrong (changes requested):
Also based on the product specification all the geometry is the same for all variants except the ID variant. In this case I recommend putting the all common geometry in meshes/irb1600 directory.

Levi-Armstrong:
@AustinDeric Friendly Ping. :)

AustinDeric:
Should we be adding external links to product specs that are out of our control?  I think we should either:
 -  host the files ourselves (easily done with our s3 bucket)
 - add the product spec to the repo

gavanderhoorn:
> host the files ourselves (easily done with our s3 bucket)

Hosting them ourselves would be a good thing. The only reason we have links to external sites is because we haven't yet implemented something else.

More generally, the problem of maintaining a repository of the source artefacts that are used when creating these support packages was something that I tried to see we could do with the [ros-industrial-source-data](https://github.com/ros-industrial-source-data) organisation. But that never really got of the ground.

Using a regular, dumb storage (compared to a versioned one) is a good idea. At least that way we can be sure that as long as we maintain that we can always retrieve our sources.

> add the product spec to the repo

we've decided not to do that in the past because it can quickly lead to rather large repositories (for some mfgs PDFs with specs can run into several MBs, one for each variant). Additionally these are chiefly binary artefacts for which versioning doesn't really make sense / work very well, and as such don't really belong in a git repository.

AustinDeric:
I separated the model specific meshes.  I added the document ID and revision to the package.xml, as discussed with Levi, this is a way to document the origin of this robot while not having to deal with broken links or hosting datasheets.

abb_irb1600_support/package.xml
```diff
@@ -8,6 +8,7 @@
     <p>
       This package contains configuration data, 3D models and launch files
       for ABB IRB 1600 manipulators. This currently includes the base model.
+      Package baed on ABB Document ID: 3HAC023604-001, Rev M.
```
Levi-Armstrong: Typo `baed` -> `based`

Levi-Armstrong (changes requested):

Levi-Armstrong:
@shaun-edwards & @gavanderhoorn Are you OK with the change in documenting the product specification document in the package.xml to prevent broken links in the future? See conversation above.

abb_irb1600_support/package.xml
AustinDeric: fixed

gavanderhoorn:
@Levi-Armstrong wrote:
> @shaun-edwards & @gavanderhoorn Are you OK with the change in documenting the product specification document in the package.xml to prevent broken links in the future? See conversation above.

in principle that is fine - I do the same for all Fanuc support packages (as those documents are not publicly available).

It would be nice though if we could make sure to also save a copy of the docs we reference, as even with a document nr it may not be trivial to get access to a copy in the future.

I think we've already have some support pkgs for which the referenced docs are no longer available.

abb_irb1600_support/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+
+if (CATKIN_ENABLE_TESTING)
+  find_package(roslaunch REQUIRED)
+  roslaunch_add_file_check(tests/roslaunch_test.xml)
+endif()
+
+foreach(dir config launch meshes urdf)
+   install(DIRECTORY ${dir}/
+      DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/${dir})
+endforeach()
```
gavanderhoorn, line 17: Could this be changed to the 'new form' (ie: single link `install(..)`)?

abb_irb1600_support/launch/load_irb1600_6_12.launch
```diff
@@ -0,0 +1,3 @@
+<launch>
+        <param name="robot_description" command="$(find xacro)/xacro --inorder '$(find abb_irb1600_support)/urdf/irb1600_6_12.xacro'" />
+</launch>
```
gavanderhoorn: indent is a tab here, should be two spaces.

abb_irb1600_support/launch/test_irb1600_6_12.launch
```diff
@@ -0,0 +1,7 @@
+<launch>
+        <include file="$(find abb_irb1600_support)/launch/load_irb1600_6_12.launch" />
+	<param name="use_gui" value="true" />
+	<node name="joint_state_publisher" pkg="joint_state_publisher" type="joint_state_publisher" />
+	<node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher" />
+	<node name="rviz" pkg="rviz" type="rviz" args="-d $(find industrial_robot_client)/config/robot_state_visualize.rviz" required="true" />
+</launch>
```
gavanderhoorn: tabs vs spaces here as well.

gavanderhoorn: And would be nice if `use_gui` could be made a private param of `joint_state_publisher`.

abb_irb1600_support/package.xml
```diff
@@ -0,0 +1,43 @@
...
+  <name>abb_irb1600_support</name>
+  <version>0.0.0</version>
+  <description>
+    <p>
+      ROS-Industrial support for the ABB IRB 1600 (and variants).
+    </p>
+    <p>
+      This package contains configuration data, 3D models and launch files
+      for ABB IRB 1600 manipulators. This currently includes the base model.
+      Package based on ABB Document ID: 3HAC023604-001, Rev M.
```
gavanderhoorn: Is the `6/12` actually the "base model"? Should we actually list the 6/12?

abb_irb1600_support/package.xml
```diff
@@ -0,0 +1,43 @@
...
+      Before using any of the configuration files and / or meshes included
+      in this package, be sure to check they are correct for the particular
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author email="aabdelmawgoud@swri.org">Ahmed Abdelmawgoud (Southwest Research Institute)</author>
+  <maintainer email="aderic@swri.org">Austin Deric (Southwest Research Institute)</maintainer>
+  <license>Apache2</license>
+
+  <url type="website">http://ros.org/wiki/abb_irb1600_support</url>
```
gavanderhoorn: This is the old url for the wiki. New one: `wiki.ros.org`.

abb_irb1600_support/urdf/irb1600_6_12.urdf
```diff
@@ -0,0 +1,174 @@
+<?xml version="1.0" ?>
```
gavanderhoorn: I don't believe we need to keep providing URDFs anymore. Xacro support is ubiquitous, and URDFs are not composible, so I suggest we remove it from this PR.

abb_irb1600_support/urdf/irb1600_6_12_macro.xacro
```diff
@@ -0,0 +1,173 @@
...
+  </link>
+  <link name="${prefix}tool0"/>
+  <!-- end of link list -->
+  <!-- joint list -->
+  <joint name="${prefix}joint_1" type="revolute">
+    <origin xyz="0 0 0" rpy="0 0 0"/>
+    <parent link="${prefix}base_link"/>
+    <child link="${prefix}link_1"/>
+    <axis xyz="0 0 1"/>
+    <limit lower="-3.1416" upper="3.1416" effort="0" velocity="2.618"/>
```
gavanderhoorn: It might be better to use `${pi}` here: `3.1416` is a rounded up version of `pi`, and I've actually had robots run into their joint limits and faulting because MoveIt used that fourth decimal in some trajectories.

Alternatively: we could use the joint limits in degrees and use `${radians(..)}`.

abb_irb1600_support/urdf/irb1600_6_12_macro.xacro
```diff
@@ -0,0 +1,173 @@
...
+    <origin xyz="0.065 0 0" rpy="0 0 0"/>
+    <parent link="${prefix}link_5"/>
+    <child link="${prefix}link_6"/>
+    <axis xyz="1 0 0"/>
+    <limit lower="-6.9813" upper="6.9813" effort="0" velocity="7.854"/>
+  </joint>
+  <joint name="${prefix}joint_6-tool0" type="fixed">
+    <parent link="${prefix}link_6"/>
+    <child link="${prefix}tool0"/>
+    <origin xyz="0 0 0" rpy="0 1.57079 0"/>
```
gavanderhoorn: Perhaps `${pi/2.0}` here instead of `1.57079`?

abb_irb1600_support/CMakeLists.txt
AustinDeric: @Levi-Armstrong can you show me what this means?  I am unfamiliar with this syntax.

abb_irb1600_support/launch/load_irb1600_6_12.launch
AustinDeric: fixed.

abb_irb1600_support/launch/test_irb1600_6_12.launch
AustinDeric: fixed x2.

abb_irb1600_support/package.xml
AustinDeric: fixed.

abb_irb1600_support/package.xml
AustinDeric: fixed.

abb_irb1600_support/urdf/irb1600_6_12.urdf
AustinDeric: fixed.

abb_irb1600_support/urdf/irb1600_6_12_macro.xacro
AustinDeric: fixed with the ${pi} solution.

abb_irb1600_support/urdf/irb1600_6_12_macro.xacro
AustinDeric: fixed with ~~${pi}/2~~ ${pi/2}

Jmeyer1292:
Assuming the code builds in travis, I think this is good enough.

+1

I'd like to see some tests on hardware prior to moving to the main repo.

PR #52, opened 2017-04-30
Rename ABB IRB4400

Jmeyer1292:
Changes robot name from `irb4400l_30` to `irb4400l_30_243` to comply with new naming conventions.

Accomplished by running the following command:
```
find . -type f | xargs sed -i 's/irb4400l_30/irb4400l_30_243/g'
```

and then manually renaming a few files.

The `243` number isn't part of the ABB designation in the manuals, so I inferred it from an old spec [here](https://library.e.abb.com/public/161ba710e316c88bc12576cb00528ef5/Product%20specification%204400%20M98A%20BWOS3.2.pdf) on page 43.


gavanderhoorn:
> The `243` number isn't part of the ABB designation in the manuals, so I inferred it from an old spec [here](https://library.e.abb.com/public/161ba710e316c88bc12576cb00528ef5/Product%20specification%204400%20M98A%20BWOS3.2.pdf) on page 43.

Seems correct. @Levi-Armstrong gave the same nrs in 2015 (in https://github.com/ros-industrial/abb_experimental/pull/21#issuecomment-77980313):

> [..] In the case of the irb 4400l there is two payloads variants 10kg and 30kg and they actual have different reaches; the 10kg is 2.55m and the 30kg is 2.43m. The suggest naming for the variants would be abb_irb4400l_10_255 and 4400l_30_243. Do you agree?

gavanderhoorn (approved):
Tested and works.

gavanderhoorn:
Thanks @Jmeyer1292. :+1:

gavanderhoorn:
Thanks for picking this up btw, @Jmeyer1292 👍

PR #59, opened 2017-05-05
Transfer IRB4400 Support Package

Jmeyer1292:
Removes the 4400 package from this repo.

See [this](https://github.com/ros-industrial/abb/pull/131) for the associated `abb` PR.


gavanderhoorn:
We should also update the [package manifest](https://github.com/ros-industrial/abb_experimental/blob/446bf69d487fc8f50904e0c45632e4c3bcf4a39e/abb_experimental/package.xml) of the metapackage to no longer include this one.

Jmeyer1292:
Updated. Also removed a non existent move it config package.

gavanderhoorn:
@Jmeyer1292 wrote:
> Also removed a non existent move it config package.

yes, I accidently added that (I had one to test the IRB4400 support pkg).

Jmeyer1292:
Good to go?

PR #58, opened 2017-05-05
[CI] Fix wrong distro.

130s:
Travis CI does not support Xenial <s>as of today</s>at the time https://github.com/travis-ci/travis-ci/issues/5821#issuecomment-243617289

Although I'm not sure if this fixes build issues like https://github.com/ros-industrial/industrial_ci/issues/167, with this change Travis passed on my forked repo.

@Jmeyer1292 @ipa-mdl

PR #44, opened 2017-04-21
fix build issues

AustinDeric:
Update package to kinetic, also:
 - fix build issues
 - fix travis
 - package version 2
 - improve readme

PR #61, opened 2017-05-08
change .travis.yml from trusty (ubuntu 14) to xenial (ubuntu 16)

AustinDeric:
I do not understand how it was building correctly before this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: xenial
 language: generic
 compiler:
   - gcc


gavanderhoorn:
@AustinDeric: I think you missed #58 and ros-industrial/industrial_ci#167.

`xenial` is not a supported value for the `dist` key.

#58 reverted #44, where this was introduced.

AustinDeric:
Whoops, sorry! 
https://github.com/ros-industrial/industrial_ci/issues/167

PR #63, opened 2017-05-26
irb4600: mention specific variant supported in manifest.

gavanderhoorn:
as per subject.

The package manifest of `abb_irb4600_support` does not make explicit which specific variant(s) it supports.

Updated it to now mention the `60/2.05` variant.


AustinDeric (approved):
+1

Thanks!

PR #64, opened 2017-05-26
Minor fixups for the IRB 52 support pkg

gavanderhoorn:
As per subject.

### DIFF
--- a/abb_irb52_support/launch/load_irb52_7_120.launch
+++ b/abb_irb52_support/launch/load_irb52_7_120.launch
@@ -1,4 +1,4 @@
 <?xml version="1.0" ?>
 <launch>
-  <param name="robot_description" command="$(find xacro)/xacro.py --inorder '$(find abb_irb52_support)/urdf/irb52_7_120.xacro'" />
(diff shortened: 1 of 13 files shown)


AustinDeric (approved):

AustinDeric:
reviewed + builds + tested = +1

I couldn't get it running in http://viz.rosindustrial.com/gavanderhoorn/abb_experimental/irb52_mention_variants/irb52_7_145

The reason is that, as stated above, we need abb_resources.  abb_resources is only material and color definitions.  Thats kinda cumbersome but this works as intended.

gavanderhoorn:
@AustinDeric wrote:
> I couldn't get it running in http://viz.rosindustrial.com/gavanderhoorn/abb_experimental/irb52_mention_variants/irb52_7_145
> 
> The reason is that, as stated above, we need abb_resources.  abb_resources is only material and color definitions.  Thats kinda cumbersome but this works as intended.

I'm not sure I understand what you mean with this.

`abb_resources` has been released into indigo quite some time ago, so should be installable via `rosdep install ..`.

PR #65, opened 2017-05-26
Mark pkgs as architecture independent where missing

gavanderhoorn:
As per subject.

### DIFF
--- a/abb_irb120_gazebo/package.xml
+++ b/abb_irb120_gazebo/package.xml
@@ -39,4 +39,7 @@
   <exec_depend>robot_state_publisher</exec_depend>
   <exec_depend>ros_controllers</exec_depend>
 
+  <export>
+    <architecture_independent/>
+  </export>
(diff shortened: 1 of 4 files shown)


AustinDeric:
Great stuff! we need to do this to alot of our packages:

https://github.com/ros-infrastructure/rep/blob/240583e7c1b5148e0dc939bd40a628fda047aa26/rep-0140.rst#architecture_independent

<architecture_independent/>

This empty tag indicates that your package produces no architecture-specific files once built. That information is intended for allowing optimization of packaging.

Specifying <architecture_independent/> is recommended for metapackages and for packages defining only ROS messages and services. Python-only packages are reasonable candidates, too. It is not appropriate for any package which compiles C or C++ code.

Be sure to remove this tag if some subsequent update adds architecture-dependent targets to a formerly independent package.

AustinDeric (approved):

PR #71, opened 2017-06-29
Add abb_irb1600_6_12_moveit_config package

keerthanamanivannan:


Levi-Armstrong:
@gavanderhoorn I think we should make moveit config packages name include the variant information. Do you agree?

gavanderhoorn:
@Levi-Armstrong: yes, I agree.

@keerthanamanivannan: could you please regenerate the MoveIt config and make it include the variant of the IRB 1600 it is for? See [REP-I0007](https://github.com/ros-industrial/rep/blob/master/rep-I0007.rst) for the naming convention.

Concretely, I think this would mean that instead of `abb_irb1600_moveit_config`, the package should be named `abb_irb1600_6_12_moveit_config`.

PS: renaming a ROS package is doable, but not trivial and it's easy to forget to update all references. Regenerating the MoveIt config should not be much more work, but is probably easier.

keerthanamanivannan:
@gavanderhoorn alright, I'm on it!

keerthanamanivannan:
@gavanderhoorn @Levi-Armstrong is this alright?

gavanderhoorn:
Naming is alright now, thanks for that @keerthanamanivannan.

I'm just wondering whether we should include 'trajectory execution support' in any MoveIt configurations that we maintain / host here. As it is right now, this MoveIt config cannot be used with a real robot yet.

We could merge it as-is and then in a next PR we could add the trajectory execution support, but I always like to do that in one go.

I'll leave it up to you and @Levi-Armstrong.

gavanderhoorn:
We would basically need to add `moveit_planning_execution.launch`, `abb_irb1600_6_12_moveit_controller_manager.launch.xml` and `controllers.yaml` with the appropriate contents.

You could take a look at the `irb120` moveit configs to see what they should contain.

gavanderhoorn:
One other suggestion: would we perhaps want to make this MoveIt config use `trac_ik` for its IK solver by default? It's currently configured to use KDL, which is not the best choice.

keerthanamanivannan:
@gavanderhoorn ok, I'll fix it!

keerthanamanivannan:
@gavanderhoorn yep, let me know if there are any more changes.

gavanderhoorn:
@keerthanamanivannan: this looks ok. Thanks for the PR :+1:.

I've pushed a minor commit that patches up `package.xml` a bit (missing dep on `trac_ik` and some other things).

We'll wait on Travis before merging.

Finally: I noticed that your commits are not being attributed to your github account (probably gmail vs company address). Would you want to fix that before I merge?

gavanderhoorn:
PS: I would also recommend to use (feature) branches in the future. Felt a little awkward pushing to your `kinetic-devel`.

keerthanamanivannan:
@gavanderhoorn haha ok, got it. feature branches from next time. Also, I changed my email address with github (it's been long pending). you can push this code as it is.

Levi-Armstrong:
@gavanderhoorn Based on the discussion it sounds like this can be merged, but wanted to check with you before merging. Is it ready?

gavanderhoorn:
@Levi-Armstrong @keerthanamanivannan: just reviewed this and I've added two small fixup commits.

Let's wait for Travis and then we can merge this.

Note: until #75 is merged two warnings will be printed by MoveIt when starting.

gavanderhoorn:
Also: let's squash-merge this to avoid merging all the fixup commits.

gavanderhoorn:
Thanks @keerthanamanivannan for the contribution :100:

PR #81, opened 2017-12-31
Add ABB IRB 1200 moveit package

stwirth:
Based off of #77 (thanks @ajshort!), added one commit to have separate files for both robot variants.
Replaces #77 (and probably #34 as well).

PR #77, opened 2017-09-15
Add ABB IRB 1200 moveit package

ajshort:

PR #85, opened 2018-01-03
IRB 1200 support pkg updates & MoveIt pkgs

gavanderhoorn:
This includes the changes contributed by @ajshort (in #77) and by @stwirth (in #81), plus a nr of smaller commits by me.

We could also opt to just merge #77 or #81 and fixup things in a subsequent PR, but the reason it might make sense not to do that is that the `7/0.7` variant can share many of its meshes with the `5/0.9`. As the `7/0.7` variant is introduced in #77 (but doesn't share any meshes with the `5/0.9` there), merging that PR would introduce commits that add (relatively) large files to the repository that would then later be removed again.

Doing it all in one PR would avoid that (as we can squash merge).


gavanderhoorn:
Opening a PR so Travis can take a look and to (if needed) discuss the proposed changes.

gavanderhoorn:
I just realised that sharing meshes between the `7/0.7` and `5/0.9` variants is difficult as one uses Collada for the `visual` meshes and the other STL.

Technically no reason those can't be mixed, but aesthetically it's a bit of a no-no.

stwirth:
Ping! I'd like to see this merged...

gavanderhoorn:
To do this properly I'll need to go and convert meshes to Collada from the meshes that ABB provides. I won't have time for that this week.

If you can provide the meshes we are missing, that would expedite things.

bhomberg:
Hi!  I'm working with @stwirth on this project.

I'm trying to figure out what exactly you need to expedite this.  As far as I can tell:
-for the 1200_5_90 arm, abb_experimental has .stl's for collision meshes
-for the 1200_5_90 arm, abb_experimental has .dae's for visual meshes
-this PR includes collision meshes and visual meshes as .stl's

Do you just need the 1200_7_70 visual meshes converted from .stl to .dae format, or do you also need some comparison of which ones are equivalent / change in file setup so they can be used together more easily?

ajshort:
If you're trying to get collada meshes, you can export them directly from robotstudio rather than converting them. Just right click on each link and go "Export Geometry".

gavanderhoorn:
@bhomberg wrote:
> I'm trying to figure out what exactly you need to expedite this

It would be nice if we could share as many meshes as possible between the two variants supported by the irb1200 package. Both packages have STL meshes for the collision quality meshes, but the `7/0.7` variant uses STL for the visual meshes and the `5/0.9` uses Collada.

Technically there is nothing preventing us from mixing the two, it just won't look very nice.

Summarising: the `7/0.7` meshes should be Collada, the overlap between the two variants determined and only a single set of meshes should be stored. See [fanuc_lrmate200ic_support](https://github.com/ros-industrial/fanuc/tree/21d16189c3640c7f161a56504edec4db26eb7735/fanuc_lrmate200ic_support/meshes) for an example (note the `lrmate200ic5l` directory which only contains meshes for `link`s 2 and 4).

gavanderhoorn:
I believe this is good to go now.

I've requested a review from @Levi-Armstrong.

I'll merge after he gives his :+1:

Levi-Armstrong (approved):

gavanderhoorn:
Thanks @Levi-Armstrong for the review.

And thanks @stwirth, @ajshort, @bhomberg and @jonbinney.

:beers: :hamburger: :+1:

gavanderhoorn:
@stwirth, @ajshort, @bhomberg and @jonbinney: I've done a squash-merge in order to avoid repository bloat due to the changes to the various meshes that were tried/created.

I've made sure to retain attribution of your contributions in the commit msg of the squash commit.

Thanks again.

jonbinney:
Thanks @gavanderhoorn !

PR #94, opened 2018-04-24
irb4600: remove 'solid' from STL header.

gavanderhoorn:
As per subject.

This avoids warnings such as:

```
The STL file 'package://abb_irb4600_support/meshes/irb4600_60_205/visual/base_link.stl' is malformed. It starts with the word 'solid', indicating that it's an ASCII STL file, but it does not contain the word 'endsolid' so it is either a malformed ASCII STL file or it is actually a binary STL file. Trying to interpret it as a binary STL file instead.
```


gavanderhoorn:
@Levi-Armstrong ?

Levi-Armstrong (approved):

gavanderhoorn:
Thanks @Levi-Armstrong.

Issue #97, opened 2018-04-25
Inertial values for IRB 1200 models?


jonbinney:
Looks like the 120 has inertial values defined in the URDF here: https://github.com/ros-industrial/abb_experimental/blob/e7e6c119c23ab257cb545450b927ed6df3920dcc/abb_irb120_support/urdf/irb120_3_58_macro.xacro

Does anyone know how those values were determined? @bhomberg and I are trying to determine those values for the IRB1200 5_90 and 7_70 models. We've contacted ABB and they aren't willing to share those values - did someone measure them?

/cc @stwirth

shaun-edwards:
I don't know where they came from, but sometimes people use the CAD model to estimate them.  As you have found, the robot vendors do not provide this information.  I would use at your own risk.

gavanderhoorn:
These were made up / guessed / approximated using SolidWorks (from ros-industrial/abb_experimental#18):

> The masses and inertial values for each link were calculated in solidworks. Adding the mass of each link gives a total of 18kg, which is lower than the robots actual value of 25kg. I contacted ABB about this, but the factory doesn't provide more accurate values for each link. They did say that solidworks values are accurate, but that they don't include the robot cable harness as the mass distribution is difficult to calculate. I updated the inertia tensors and cog origins to account for the new joint translations, but it might be worth double checking these values. The inertia tensor for link 6 seemed was causing problems in Gazebo (values were too small) so I ended up using a simplified matrix for this link.

See also https://github.com/ros-industrial/kuka_experimental/issues/128#issuecomment-383111315.

gavanderhoorn:
I'm going to close this as it's not really an issue with the packages.

Feel free to keep commenting though.

jonbinney:
Thanks for the info! We'll estimate the intertias as well as we can and submit a PR. They'll be useful for doing more realistic gazebo simulation of the robot.

PR #98, opened 2018-04-27
Gazebo package for the abb_irb1200

stwirth:
Adds a gazebo package for the abb_irb1200. For now this package only contains files for the 5_90 variant.
I had to add inertial values to the robot's URDF in the abb_irb1200_support package to make gazebo happy. Those values are not accurate though, they are stolen from abb_irb120_gazebo (see also https://github.com/ros-industrial/abb_experimental/issues/97).

image

You can test this running:
```
$ roslaunch abb_irb1200_gazebo irb1200_5_90_gazebo.launch
$ rosrun rqt_joint_trajectory_controller rqt_joint_trajectory_controller arm_controller/command:=joint_path_command arm_controller/state:=feedback_states
```

### DIFF
--- a/abb_irb1200_gazebo/CMakeLists.txt
+++ b/abb_irb1200_gazebo/CMakeLists.txt
@@ -1,0 +1,12 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(abb_irb1200_gazebo)
+
+find_package(catkin REQUIRED)
+
+catkin_package()
+
+foreach(dir config launch urdf)
+ install(DIRECTORY ${dir}/
+         DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/${dir})
+endforeach()
+
--- a/abb_irb1200_gazebo/config/irb1200_5_90_arm_controller.yaml
+++ b/abb_irb1200_gazebo/config/irb1200_5_90_arm_controller.yaml
@@ -1,0 +1,21 @@
+arm_controller:
+  type: position_controllers/JointTrajectoryController
+  joints:
+     - joint_1
+     - joint_2
+     - joint_3
+     - joint_4
+     - joint_5
+     - joint_6
+  constraints:
+      goal_time: 0.6
+      stopped_velocity_tolerance: 0.05
+      joint_1: {trajectory: 0.1, goal: 0.1}
+      joint_2: {trajectory: 0.1, goal: 0.1}
+      joint_3: {trajectory: 0.1, goal: 0.1}
+      joint_4: {trajectory: 0.1, goal: 0.1}
+      joint_5: {trajectory: 0.1, goal: 0.1}
+      joint_6: {trajectory: 0.1, goal: 0.1}
+  stop_trajectory_duration: 0.5
+  state_publish_rate:  25
+  action_monitor_rate: 10
--- a/abb_irb1200_gazebo/config/joint_state_controller.yaml
+++ b/abb_irb1200_gazebo/config/joint_state_controller.yaml
@@ -1,0 +1,3 @@
+joint_state_controller:
+  type: joint_state_controller/JointStateController
+  publish_rate: 50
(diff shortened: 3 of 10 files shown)


gavanderhoorn:
Thanks for this.

I'm not sure how comfortable I am with copying the dynamic params from the IRB 120 though.

How about using cylindrical approximations, like is done in the UR gazebo xacros [here](https://github.com/ros-industrial/universal_robot/blob/ad740e0dd028cd9614b869d306dfc2da402cbc46/ur_description/urdf/ur10.urdf.xacro#L53-L61)? We could add that macro to a new file in `abb_resources` and make use of it in the Gazebo macro for the IRB 1200.

The approximations are not going to be too accurate, but it'll probably be better than copying them from a much smaller and lighter robot.

gavanderhoorn:
@stwirth: is this something you're still using?

I'm still not sure whether I'd be comfortable with copying the values of the IRB 120.

Could you comment on your experiences with this setup?

stwirth:
@gavanderhoorn I'm not on the project anymore. @bhomberg might want to take this over.

gavanderhoorn:
Ok, thanks for commenting.

@bhomberg: can you comment on the behaviour of the proposed simulation?

jonbinney:
I can spend some time verifying the inertial values and testing behavior in simulation.

jonbinney:
Hmmm - I found this paper: https://www.researchgate.net/publication/324960713_Energy_consumption_modeling_of_industrial_robot_based_on_simulated_power_data_and_parameter_identification

They did parameter identification for an IRB 1200 and provide the estimated values in table 2. I'll compare it to what we guessed.

gavanderhoorn:
> Hmmm - I found this paper: https://www.researchgate.net/publication/324960713_Energy_consumption_modeling_of_industrial_robot_based_on_simulated_power_data_and_parameter_identification
> 
> They did parameter identification for an IRB 1200 and provide the estimated values in table 2. I'll compare it to what we guessed.

Awesome. That would be even better of course.

Thanks Jon.

gavanderhoorn:
We might also want to consider using effort joint interfaces instead of position ones.

Previously we've always used position control for Gazebo simulations, but that is inherently unstable when interacting with other objects in Gazebo and introduces quite a few problems.

The downside is that we'd need to have proper inertias and masses -- which we now seem to have for this model at least -- and we'd need to tune the PIDs.

That is unfortunate, but all-in-all going with effort control will most likely greatly increase the utility of the package.

gavanderhoorn:
@jonbinney: have you had any chance to look at those inertia parameters you found?

jonbinney:
The inertia values are expressed in terms of "regrouped" parameters.  Here are the results they got for the irb1200 5/90:
screenshot from 2019-01-13 19-26-54

The m parameters are friction params that we don't need. The I parameters are what we want, but to decode what they've put in R they say only: "The method for minimum set of inertial parameters can be calculated by Gautier and Khalil.24" The refernced paper is:

 "Gautier M and Khalil W. A direct determination of minimum inertial parameters of robots. In: Proceedings of the 1988 IEEE international conference on robotics and automation, Philadelphia, PA, 24–29 April 1988, vol. 3, pp.1682–1687. New York: IEEE."

Which seems to be this: https://www.computer.org/csdl/proceedings/robot/1988/0852/00/00012308.pdf

The core idea seems to be that it isn't feasible to estimate the parameters for some links because you can't move them in a way that will make certain dimensions of the inertia observable. For example we can only rotate link 1 around Z, and so we can only estimate the ZZ component of the inertia. So, unfortunately, the results of this paper aren't useful for setting all params of the inertia, which we need to to in the URDF format.

jonbinney:
You can go ahead and close this for the moment - I will reopen if I have time to approximate the params in some other way (e.g. cylinders of constant density)

gavanderhoorn:
Closing as we have #110 as a replacement.

Thanks @stwirth for the initial work!

PR #100, opened 2018-07-11
update and unify Readme files

NadiaHG:


gavanderhoorn:
We're going to do this slightly differently in this repository.

Thanks for the initial work @ipa-nhg :+1:

PR #110, opened 2019-03-05
Add gazebo support for IRB1200 5/90 variant

tulku:
* Adds the irb1200_gazebo package with support for the 5 90 variant using effort controllers.
* Adds inertial values to abb_irb1200_support irb1200_5_90 xacro file

This is built upon the initial work of @stwirth (https://github.com/ros-industrial/abb_experimental/pull/98) and @jonbinney

## Intertials

Meshlab was used to calculate the inertia tensor, the center of mass and the volume of each collision mesh. This is the method explained in this [very well known tutorial](http://gazebosim.org/tutorials?tut=inertia). Meshes were scaled 10 times to get better results.

To calculate each link mass a constant density in the entire robot was assumed and then distributed the mass using the calculated volume of each link.

I used the obtained values for all joints except for `joint_6` where they made the robot "crash". Maybe it's because the values are too small?

The new inertials look like this:
screenshot from 2019-02-28 19-56-24

I tested the cylinder approximation for link 6 but those values failed in the same way, and were also very small.

## Dynamics and control

Gazebo plotting tool was use to get and idea of the robot responses to motion commands. The original values made the robot oscillate significantly while moving, as can be seen on this plot:
screenshot from 2019-02-28 18-58-01

The new values make the robot move smoothly and pretty quickly:
screenshot from 2019-02-28 19-43-44

I'm pretty sure that we can find even better values if we keep tuning, for example to reduce even more the steady state errors, but I believe that the proposed values are already very adequate.

## Testing

Start the simulation:
```
$ roslaunch abb_irb1200_gazebo  irb1200_5_90_gazebo.launch
```
Test the controllers:
```
$ rosrun rqt_joint_trajectory_controller rqt_joint_trajectory_controller arm_controller/command:=joint_path_command arm_controller/state:=feedback_states
```

### DIFF
--- a/abb_irb1200_gazebo/package.xml
+++ b/abb_irb1200_gazebo/package.xml
@@ -44,7 +44,6 @@
   <exec_depend>joint_trajectory_controller</exec_depend>
   <exec_depend>position_controllers</exec_depend>
   <exec_depend>robot_state_publisher</exec_depend>
-  <exec_depend>ros_controllers</exec_depend>
 
   <export>
     <architecture_independent/>
--- a/abb_irb1200_gazebo/package.xml
+++ b/abb_irb1200_gazebo/package.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>abb_irb1200_gazebo</name>
+  <version>0.0.1</version>
+  <description>
+    <p>
+      Gazebo support for the ABB IRB 1200
+    </p>
+    <p>
+      This package contains configuration and launch files
+      to simulate the ABB IRB 1200 with Gazebo.
+    </p>
+  </description>
+
+  <author email="stephan@ironox.com">Stephan Wirth</author>
+  <maintainer email="lucas@ironox.com">Lucas Chiesa</maintainer>
+
+  <license>Apache 2.0</license>
+
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <exec_depend>abb_irb1200_support</exec_depend>
+  <exec_depend>gazebo_ros_control</exec_depend>
+  <exec_depend>gazebo_ros_pkgs</exec_depend>
+  <exec_depend>gazebo_ros</exec_depend>
+  <exec_depend>gazebo</exec_depend>
+  <exec_depend>joint_state_controller</exec_depend>
+  <exec_depend>joint_trajectory_controller</exec_depend>
+  <exec_depend>position_controllers</exec_depend>
+  <exec_depend>robot_state_publisher</exec_depend>
+  <exec_depend>ros_controllers</exec_depend>
+
+  <export>
+  </export>
+</package>
--- a/abb_irb1200_gazebo/package.xml
+++ b/abb_irb1200_gazebo/package.xml
@@ -4,36 +4,49 @@
   <version>0.0.1</version>
   <description>
     <p>
-      Gazebo support for the ABB IRB 1200
+      ROS-Industrial Gazebo support package for the ABB IRB 1200 (and variants).
     </p>
     <p>
-      This package currently only contains configuration and launch files to
-      simulate the 5/0.9 variant of the ABB IRB 1200 robot.
+      This package contains the configuration data and launch files required
+      to simulate the ABB IRB 1200 manipulator in Gazebo. This currently only
+      includes the 5/0.9 variant.
     </p>
     <p>
-      Ros control <em>effort_controllers</em> are used to move the robot joints.
-      The controller params (PID, constrains) are guesstimates, and were
+      `ros_control` `effort_controllers` are used to move the robot joints.
+      The controller parameters (PID, constrains) are guesstimates, and were
       obtained testing different values until the joint movement was good enough
       for the intended application. You may need to change them for your simulation.
     </p>
+    <p>
+      Before using any of the configuration files included in this package, be
+      sure to check they are correct for the particular robot model and
+      configuration you intend to use them with.
+    </p>
   </description>
-
   <author email="stephan@ironox.com">Stephan Wirth</author>
   <maintainer email="lucas@ironox.com">Lucas Chiesa</maintainer>
-
+  <maintainer email="levi.armstrong@swri.org">Levi Armstrong (Southwest Research Institute)</maintainer>
+  <maintainer email="g.a.vanderhoorn@tudelft.nl">G.A. vd. Hoorn (TU Delft Robotics Institute)</maintainer>
   <license>Apache 2.0</license>
+
+  <url type="website">http://wiki.ros.org/abb</url>
+  <url type="bugtracker">https://github.com/ros-industrial/abb_experimental/issues</url>
+  <url type="repository">https://github.com/ros-industrial/abb_experimental</url>
 
   <buildtool_depend>catkin</buildtool_depend>
 
   <exec_depend>abb_irb1200_support</exec_depend>
+  <exec_depend>gazebo</exec_depend>
+  <exec_depend>gazebo_ros</exec_depend>
   <exec_depend>gazebo_ros_control</exec_depend>
   <exec_depend>gazebo_ros_pkgs</exec_depend>
-  <exec_depend>gazebo_ros</exec_depend>
-  <exec_depend>gazebo</exec_depend>
   <exec_depend>joint_state_controller</exec_depend>
   <exec_depend>joint_trajectory_controller</exec_depend>
   <exec_depend>position_controllers</exec_depend>
   <exec_depend>robot_state_publisher</exec_depend>
   <exec_depend>ros_controllers</exec_depend>
 
+  <export>
+    <architecture_independent/>
+  </export>
 </package>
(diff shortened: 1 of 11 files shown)


abb_irb1200_gazebo/package.xml
```diff
@@ -0,0 +1,35 @@
...
+<package format="2">
+  <name>abb_irb1200_gazebo</name>
+  <version>0.0.1</version>
+  <description>
+    <p>
+      Gazebo support for the ABB IRB 1200
+    </p>
+    <p>
+      This package contains configuration and launch files
+      to simulate the ABB IRB 1200 with Gazebo.
```
gavanderhoorn: Could you please mention the specific variant for which you added Gazebo support?

abb_irb1200_gazebo/package.xml
```diff
@@ -0,0 +1,35 @@
...
+  <exec_depend>gazebo_ros</exec_depend>
+  <exec_depend>gazebo</exec_depend>
+  <exec_depend>joint_state_controller</exec_depend>
+  <exec_depend>joint_trajectory_controller</exec_depend>
+  <exec_depend>position_controllers</exec_depend>
+  <exec_depend>robot_state_publisher</exec_depend>
+  <exec_depend>ros_controllers</exec_depend>
+
+  <export>
+  </export>
```
gavanderhoorn, line 50: Please remove.

abb_irb1200_gazebo/urdf/irb1200_5_90.xacro
```diff
@@ -0,0 +1,28 @@
...
+
+<robot name="abb_irb1200_5_90" xmlns:xacro="http://ros.org/wiki/xacro">
+  <xacro:include filename="$(find abb_irb1200_gazebo)/urdf/irb1200_5_90_macro.xacro" />
+  <xacro:include filename="$(find abb_irb1200_support)/urdf/irb1200_5_90_macro.xacro" />
+
+  <!-- get base ABB IRB1200 model -->
+  <xacro:abb_irb1200_5_90 prefix="" />
+
+  <!-- get gazebo add-ons -->
+  <xacro:abb_irb1200_5_90_gazebo_spec prefix="" />
```
gavanderhoorn: The IRB 120 Gazebo pkg includes the Gazebo/`ros_control` specific bits in this file (see [abb_irb120_gazebo/irb120_3_58_macro.xacro](https://github.com/ros-industrial/abb_experimental/blob/f8e2bf903c0875b0799521ec383d15b76344c249/abb_irb120_gazebo/urdf/irb120_3_58_macro.xacro), with the `world` link bits being added by the [top-level xacro](https://github.com/ros-industrial/abb_experimental/blob/f8e2bf903c0875b0799521ec383d15b76344c249/abb_irb120_gazebo/urdf/irb120_3_58.xacro)).

Would it make sense to do that here as well?

Did you have a specific reason for separating it all out into the `abb_irb1200_5_90_gazebo_spec` macro?

abb_irb1200_gazebo/urdf/irb1200_5_90_macro.xacro
```diff
@@ -0,0 +1,86 @@
...
+      </joint>
+      <actuator name="${prefix}motor6">
+        <mechanicalReduction>1</mechanicalReduction>
+      </actuator>
+    </transmission>
+    <!-- end of transmission list -->
+
+    <!-- Gazebo-specific link properties -->
+    <gazebo reference="${prefix}base_link">
+      <turnGravityOff>true</turnGravityOff>
```
gavanderhoorn, line 68: Why is gravity turned off here?

abb_irb1200_support/urdf/irb1200_5_90_macro.xacro
```diff
@@ -104,42 +145,48 @@
       <axis xyz="0 0 1"/>
       <parent link="${prefix}base_link"/>
       <child link="${prefix}link_1"/>
-      <limit effort="0" lower="-2.967" upper="2.967" velocity="5.027"/>
+      <limit effort="1000" lower="-2.967" upper="2.967" velocity="5.027"/>
```
gavanderhoorn, line 156: If we don't know the effort limits, we usually keep them at `0`. Is that not possible?

abb_irb1200_gazebo/package.xml
gavanderhoorn: Also: I believe it would be good to document *where* the dynamic properties came from. Especially since these are guestimates (however nice they are), it would be good for future users to understand how you arrived at them to be able to gauge their usability in certain (their) contexts.

gavanderhoorn (changes requested):

gavanderhoorn:
Thanks for the PR @tulku :+1:.

I've left some questions and comments.

gavanderhoorn:
@tulku: would it be possible to address the requested changes?

tulku:
@gavanderhoorn I need to take care of other things now, but I'd like to do them when I have time.

abb_irb1200_gazebo/urdf/irb1200_5_90.xacro
tulku: Not really, everything is now in the `_gazebo` macro (removed the `_spec` also, as other robots don't have it)

abb_irb1200_gazebo/urdf/irb1200_5_90_macro.xacro
tulku: Gazebo support for the irb1200 was based off the irb120 package which also [disables gravity](https://github.com/ros-industrial/abb_experimental/blob/ce65133979980bab79a4f8f8c3afe78c4eb126c7/abb_irb120_gazebo/urdf/irb120_3_58_macro.xacro#L67-L94) for the arm links.

I now tried to make the model work without those lines (I agree that it would be better not to have them), without success. I needed to increase the friction and in general the forces applied which made the controller sometimes unstable and not perform very well. I would leave it as is.

abb_irb1200_support/urdf/irb1200_5_90_macro.xacro
tulku: It seems that when using `effort_controllers` the max effort specified in the URDF is used to limit the joint effort. Making it `0`, does not seem to mean that there is no limit, rather, it means that the joint cannot do any effort.

Trying to trace a bit the code to justify this I found [where the limits are loaded](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/2901fe6cc16c86a68cb01d664a45276cac9ec712/gazebo_ros_control/src/default_robot_hw_sim.cpp#L393) and that the limits [are enforced](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/2901fe6cc16c86a68cb01d664a45276cac9ec712/gazebo_ros_control/src/default_robot_hw_sim.cpp#L302-L303) when 'writing' to the sim, which results on [all joint commands being 0](https://github.com/ros-controls/ros_control/blob/f38bf6d7c21e744e37b254d86f1ab7796ea803f7/joint_limits_interface/include/joint_limits_interface/joint_limits_interface.h#L309).

So, to the best of my knowledge, we cannot leave `effort=0` if we want to use effort controllers (instead of position controllers as used [in other packages](https://github.com/ros-industrial/abb_experimental/blob/ce65133979980bab79a4f8f8c3afe78c4eb126c7/abb_irb120_gazebo/config/irb120_3_58_arm_controller.yaml#L2)).

abb_irb1200_gazebo/package.xml
tulku: I added some information to the `package.xml` files. Let me know if that's not enough.

tulku:
@gavanderhoorn I've updated this PR addressing some of your comments. I can rebase on kinetic-devel when its ready to merge.

gavanderhoorn:
@tulku: apologies for taking so long, but I've got a few commits I'd like to add to this PR.

Minor changes to the package manifest and `.macro_xacro` only.

I don't believe you've checked "Allow edits from maintainers", so I can't push them directly.

Would you be willing to enable that? I believe you can still click the checkmark. It should be on the right side of the first post in this PR (see also [Allowing changes to a pull request branch created from a fork](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork)).

If not I can open a PR against your PR.

tulku:
@gavanderhoorn thanks for looking into this PR!

The "Allow edits from maintainers" was indeed marked. I unmarked and marked it again to see if that helps... If it does not work, I can quickly merge a PR against my branch.

gavanderhoorn:
Just tried again and I can't push to `feature/abb_irb1200_5_90_gazebo` on your fork.

Do you have branch protection configured?

tulku:
@gavanderhoorn No, I don't see any branch protection configured. I just invited you as collaborator to the fork, with that you should be able to push.

gavanderhoorn (approved):
Looks good like this.

gavanderhoorn:
Thanks again for the PR @tulku :+1: 

performance is indeed quite nice with these gains and dynamics properties.

Thanks for taking the time to contribute it.

gavanderhoorn:
Squash merging as all commits but the first one are fixups or review changes.

PR #103, opened 2018-10-30
Added support for IRB4600 20/2.50 and 40/2.55

rtonnaer:
this adds support for the IRB4600 20/2.50 and 40/2.55 variants. these variants only have minor differences from the IRB4600 that was already supported.

PR #112, opened 2019-03-21
irb4600: add 20/2.50 and 40/2.55 variants

rtonnaer:
Replacement for #103.


RyanDelGizzi:
Hello all,

I can successfully launch the "demo.launch" file from the moveit package I created.  However, launching my full simulation in gazebo fails because there is no mass or inertia specified for the links. The error is explained in [this post](https://answers.ros.org/question/238899/this-robot-has-a-joint-named-joint_1-which-is-not-in-the-gazebo-model/).  So long as there are values (not necessarily the correct ones) gazebo runs correctly because of the controller being used.

I'll let @gavanderhoorn  advise on the best path forward.

gavanderhoorn:
Gazebo compatibility can be added in a later PR.

For now I believe this is an OK contribution.

I'll take a look at the mesh fidelity vs file size, as I believe that could be tweaked a bit (especially for `.stl`s).

RyanDelGizzi:
Is there any chance this will get reviewed and merged soon?

rtonnaer:
@RyanDelGizzi I think Gijs was delaying the merge because of the size of the meshes. I have now reduced the mesh sizes as much as possible.

@gavanderhoorn can you review my latest commits. I am not capable of reducing the mesh sizes any further, without compromising the quality.

gavanderhoorn:
@rtonnaer: I've pushed a commit that fixes the orientation (in X dimension) of `link_4` on both variants contributed in this PR.

They appeared to be off by 90 degrees (probably happened during export after decimation the meshes).

gavanderhoorn:
We'll wait for Travis and then I'll squash-merge this PR.

Most commits are fixups and they don't need to end up in the history of this repository.

gavanderhoorn (dismissed):

gavanderhoorn:
I won't actually squash-merge this, but I also don't want to merge all these commits, so I'll submit a condensed / rearranged version of this PR tomorrow.

gonzalocasas:
@gavanderhoorn et al: how is this looking? do you need any assistance to get this merged? we have exactly this same variant of irb4600 and were about to contribute it, but noticed it was already in the pipeline for merging :)

gavanderhoorn:
Replaced by #117.

gavanderhoorn:
Thanks for iterating so long @rtonnaer. Much appreciated.

See my replacement PR for the same changes, but with a bit cleaner commit history.

PR #119, opened 2020-03-11
Add flange frame (REP199)

dave992:
The support packages in `abb_experimental` do not have `flange` frames as suggested in [`REP 199 (Draft)`](http://gavanderhoorn.github.io/rep/rep-0199.html) which can be used to attach end-effectors to. Attaching end-effectors to the `tool0` frame results in additional rotation of end-effector joints if following [`REP 103`](https://www.ros.org/reps/rep-0103.html) or inconsistent frame orientations if this is ignored. 

I have made changes to all macro files in the support packages of `abb_experimental` to add the `flange` frame resulting in the following frame hierarchy:

    base_link
    ├ base
    └ link_1
        └ ..
        └ link_N
         └ flange
           └ tool0

I've tested the macro files by inspecting the TF tree and running the test launch files: 
    `roslaunch abb_irbXXXX_support test_irbXXXX_X_X.launch`

As there were also some `moveit_config` packages, I have also tested `demo.launch` of those MoveIt packages to make sure the configuration still works.


dave992:
The `abb_irb120_support` package contains additional `.urdf` files that were autogenerated. What is the reason of these files being in the repository? I have not updated these yet as they felt out of place. 

Let me know if there is a specific reason for this and if I should updated the generated `.urdf` files for this package.

dave992 (commented):
This `irb1600_6_12_macro.xacro` was also missing the `base` frame, added that as well

abb_irb120_support/urdf/irb120_3_58_macro.xacro
```diff
@@ -203,19 +202,30 @@
...
+    <joint type="fixed" name="${prefix}joint_6-flange">
+      <origin xyz="0 0 0" rpy="0 0 0"/>
+      <parent link="${prefix}link_6"/>
+      <child link="${prefix}flange"/>
+    </joint>
+
+    <!-- ROS-Industrial 'tool0' frame: all-zeros tool frame -->
+    <link name="${prefix}tool0" />
+    <joint name="${prefix}link_6-tool0" type="fixed">
+      <origin xyz="0 0 0" rpy="0 ${radians(0)} 0" />
```
gavanderhoorn: I haven't checked yet, but is this `0` correct here?

abb_irb7600_support/urdf/irb7600_150_350_macro.xacro
```diff
@@ -184,24 +183,36 @@
...
+    <joint type="fixed" name="${prefix}joint_6-flange">
+      <origin xyz="0 0 0" rpy="0 0 0"/>
+      <parent link="${prefix}link_6"/>
+      <child link="${prefix}flange"/>
+    </joint>
+
+    <!-- ROS-Industrial 'tool0' frame: all-zeros tool frame -->
+    <link name="${prefix}tool0" />
+    <joint name="${prefix}link_6-tool0" type="fixed">
+      <origin xyz="0 0 0" rpy="0 ${pi/2.0} 0" />
```
gavanderhoorn: `radians(..)` here as well?

gavanderhoorn:
Thanks for the PR! Much appreciated. :+1: 

And just to be extra clear: the REP referred to is not yet an accepted or active REP. But the `flange` frame part of it is something we've been essentially using `tool0` for, for a long time.

gavanderhoorn:
> The `abb_irb120_support` package contains additional `.urdf` files that were autogenerated. What is the reason of these files being in the repository? I have not updated these yet as they felt out of place.

No need to update the `.urdf`s. We should remove those.

In fact, please do so with this PR, to make sure we keep things consistent.

---

Edit: let's make sure they are not loaded by the accompanying MoveIt configuration packages btw.

abb_irb120_support/urdf/irb120_3_58_macro.xacro
dave992: Have the same question actually. I just copied the original value in for now but will check in RobotStudio to make sure this is correct. Forgot to mention that in the PR. The same applies to `irb120t_3_58_macro.xacro`

abb_irb7600_support/urdf/irb7600_150_350_macro.xacro
dave992: Changed it to use the `radians(..)` function (also for other models that I missed) in commit https://github.com/ros-industrial/abb_experimental/pull/119/commits/1635a93b5ef8c576ecd5abc32dac115906b9c7d8

dave992:
> > The `abb_irb120_support` package contains additional `.urdf` files that were autogenerated. What is the reason of these files being in the repository? I have not updated these yet as they felt out of place.
> 
> No need to update the `.urdf`s. We should remove those.
> 
> In fact, please do so with this PR, to make sure we keep things consistent.
> 
> Edit: let's make sure they are not loaded by the accompanying MoveIt configuration packages btw.

I have deleted the `.urdf` files (https://github.com/ros-industrial/abb_experimental/pull/119/commits/fb612fe1c4318c8709ab2af709077fd65579e76a) and replaced all references to it in the MoveIt configuration files to instead use the `.xacro` files (https://github.com/ros-industrial/abb_experimental/pull/119/commits/adb963033fe7ec2072054518c2abafd377dae8e8). 

Tested the changes by running the `demo.launch` files of the `abb_irb120*_moveit_config` packages.

abb_irb120_support/urdf/irb120_3_58_macro.xacro
gavanderhoorn: Have you had a chance to check this? I'm pretty sure this is incorrect.

However, we'll fix it in a later PR.

gavanderhoorn:
One request: could we perhaps not refer to "REP 199" in the commit comments?

That nr is not reserved or special; I just picked it as it didn't collide with existing/active REPs.

If/when there is an actual REP-199, the reference in the commit comments won't make sense any more (because either a different REP will have that nr, or the REP we're now referring to with `199` will have a different nr).

dave992:
> One request: could we perhaps not refer to "REP 199" in the commit comments?
> 
> That nr is not reserved or special; I just picked it as it didn't collide with existing/active REPs.
> 
> If/when there is an actual REP-199, the reference in the commit comments won't make sense any more (because either a different REP will have that nr, or the REP we're now referring to with `199` will have a different nr).

I have removed the reference to REP 199 in the commit messages

abb_irb120_support/urdf/irb120_3_58_macro.xacro
dave992: Checked it just now and it was indeed an error. Added a commit to fix the error (https://github.com/ros-industrial/abb_experimental/pull/119/commits/0ae97e142ea6828c1f928f8525b6a9abc5635782)

gavanderhoorn (approved):
Thanks for the PR. :+1:

gavanderhoorn:
And thanks for fixing the commit comments.

PR #120, opened 2020-03-12
Fix incorrect driver launch file include in IRB 1200 moveit configs

### LINKED ISSUE #107: A bug in abb_irb1200_5_90_moveit_config/launch/moveit_planning_execution.launch
The code is:
 run the "real robot" interface nodes 

```xml
    include file="$(find abb_irb1200_support)/launch/robot_interface_streaming_irb1200_5_90.launch" 
```

But the file `robot_interface_streaming_irb1200_5_90.launch` is not exists in `abb_irb1200_support/launch`.

We should replace it with the following：

```xml
   include file="$(find abb_irb1200_support)/launch/robot_interface_download_irb1200_5_90.launch" 
```

QAQ Sorry,I don't know how to add a piece of code, so I use the form of text.

gavanderhoorn:
As per subject.

Fixes #107.

### DIFF
--- a/abb_irb1200_5_90_moveit_config/launch/moveit_planning_execution.launch
+++ b/abb_irb1200_5_90_moveit_config/launch/moveit_planning_execution.launch
@@ -33,7 +33,7 @@
   <!--   - this typically includes: robot_state, motion_interface, and joint_trajectory_action nodes -->
   <!--   - replace these calls with appropriate robot-specific calls or launch files -->
   <group unless="$(arg sim)">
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
This went unnoticed for so long because of missing tests for these `.launch` files (`roslaunch_tests` have become impossible, as `warehouse_ros` doesn't exist any more, causing tests to fail with unrelated failures).

Levi-Armstrong (approved):

gavanderhoorn:
Thanks @Levi-Armstrong.

PR #117, opened 2020-01-15
Cleaned up PR 112 (add 20/2.50 and 40/2.55 variants to IRB 4600)

gavanderhoorn:
As per subject.

This is a cleaned up version of #112, which had too many commits and a very complex commit history.

I've kept provenance on all commits.

Locally tested, seems to be working.

PR #126, opened 2020-06-15
Fix length of link5 in IRB4600-20/2.50

simonschmeisser:
According to https://library.e.abb.com/public/fcaa8b25ff844fe5a115aa00b5a1578d/3HAC032885%20PS%20IRB%204600-en.pdf  there is a `0` missing

Seems to be a good idea to write opw configs based on original specifications


gavanderhoorn (approved):
Not sure how that slipped in the review of #117.

:+1:

gavanderhoorn (changes requested):
Could I please ask you to update the commit comment with a reference to the document id and revision? So not a URL.

simonschmeisser:
it's the same as in the package.xml but sure I can add it

gavanderhoorn (approved):

gavanderhoorn:
Thanks @simonschmeisser

PR #136, opened 2022-06-29
Add ABB IRB 1600 8/1.45 variant

gonzalocasas:
This PR adds the ABB IRB 1600 8/1.45 variant to the `abb_irb1600_support` package. If I understand it correctly, it is no longer desirable to include the moveit_config package that goes with it, so I have skipped this part.

The difference between this new variant and the existing IRB 1600 is on the link2 and the limits of all joints.

Please let me know if any changes are required.

PR #137, opened 2022-06-29
add IRB1600 10/1.45, add opw config for IRB1600 10/1.45, 8/1.45 and 6/1.2

cjue:
Update reference document in package.xml

Add mesh for the 1.45 variant of link2, with scale 0.001 and transform -.150, -0.4865, -115

Rename files fo irb1600 6/1.2 for clarity.


gavanderhoorn:
> Add mesh for the 1.45 variant of link2, with scale 0.001 and transform -.150, -0.4865, -115

have you since updated the PR? I don't see this scaling.

(if it was there, I would've asked you to scale the `.stl` itself though)

> Rename files fo irb1600 6/1.2 for clarity.

only the `roslaunch_test` `.xml`, correct?

abb_irb1600_support/urdf/irb1600_10_145_macro.xacro
```diff
@@ -0,0 +1,191 @@
...
+  <!-- link list -->
+  <link name="${prefix}base_link">
+    <visual>
+      <origin xyz="0 0 0" rpy="0 0 0"/>
+      <geometry>
+        <mesh filename="package://abb_irb1600_support/meshes/irb1600/visual/base_link.stl"/>
+      </geometry>
+      <material name="abb_orange">
+        <color rgba="1 0.43 0 1"/>
+      </material>
```
gavanderhoorn: could you please use the shared colour / material definitions from `abb_resources` instead?

(same for all the other occurrences below)

abb_irb1600_support/urdf/irb1600_10_145_macro.xacro
```diff
@@ -0,0 +1,191 @@
...
+    </collision>
+  </link>
+  <!-- end of link list -->
+  <!-- joint list -->
+  <joint name="${prefix}joint_1" type="revolute">
+    <origin xyz="0 0 0" rpy="0 0 0"/>
+    <parent link="${prefix}base_link"/>
+    <child link="${prefix}link_1"/>
+    <axis xyz="0 0 1"/>
+    <limit lower="${radians(-180)}" upper="${radians(180)}" effort="0" velocity="${radians(180)}"/>
```
gavanderhoorn, line 124: Thanks for using `radians(..)` here.

The other variants don't use it yet, but it's good to start converting when adding new variants.

gavanderhoorn (changes requested):
Thanks for the PR :+1: 

I've left one in-line comment. Could you take a look?

gavanderhoorn:
Friendly ping @cjue ?

cjue:
I just noticed that #136 has introduced the irb1600 8/1.45 in the meantime.

Interesting tidbit: the IRB1600 10/1.45 is a re-release / "upgrade" of the IRB1600 8/1.45 that's identical except for payload and velocity limits.

> Revision U: "IRB 1600-8 upgraded to 10 kg"
> 
Source: https://search.abb.com/library/Download.aspx?DocumentID=3HAC023604-001&LanguageCode=en&

I updated both 1.45 urdfs to use the same link_2 meshes.

abb_irb1600_support/urdf/irb1600_10_145_macro.xacro
cjue: done, also for the 6/1.2 variant.

cjue:
> > Add mesh for the 1.45 variant of link2, with scale 0.001 and transform -.150, -0.4865, -115
> 
> have you since updated the PR? I don't see this scaling.
> 
> (if it was there, I would've asked you to scale the `.stl` itself though)

I meant to say this was how I got from the STEP file to the mesh. It's also wrong, as I found out.

When comparing my link_2 mesh with the 8/1.45 version I noticed that I had set the origin incorrectly, placing link_2 inside the neighboring joints. This PR now uses the correct mesh for both the 8/1.45 and the 10/1.45.


> > Rename files fo irb1600 6/1.2 for clarity.
> 
> only the `roslaunch_test` `.xml`, correct?

Yes, the other files already had the model name in there.

cjue:
Just a note: This PR does not touch the `J23_coupled` settings, but #139 will affect all IRB1600 models.

cjue:
closing and reopening to nudge checks

cjue:
> Friendly ping @cjue ?

ping addressed @gavanderhoorn

gavanderhoorn:
Closing in favour of #154.

Thanks for this PR @cjue :+1: 

and apologies for letting it sit for so long.

PR #138, opened 2022-06-29
Use opw_kinematics instead of trac-ik for IRB1200

cjue:


abb_irb1200_5_90_moveit_config/config/kinematics.yaml
```diff
@@ -1,5 +1,12 @@
...
+  kinematics_solver: moveit_opw_kinematics_plugin/MoveItOPWKinematicsPlugin
+  opw_kinematics_geometric_parameters:
+      a1:  0
+      a2: -0.042
+      b:   0.0
+      c1:  0.3991
+      c2:  0.448
+      c3:  0.451
+      c4:  0.082
+  opw_kinematics_joint_offsets: [0.0, 0.0, -1.57079632679, 0.0, 0.0, 0.0]
```
gavanderhoorn: Any reason not to use `deg(..)` here?

gavanderhoorn (dismissed):
Looks OK to me.

I've left one in-line comment.

If you can confirm you've verified these parameters are correct, we can merge.

abb_irb1200_5_90_moveit_config/config/kinematics.yaml
cjue: No, thanks for the reminder, added a commit with `deg()` syntax.

cjue:
> If you can confirm you've verified these parameters are correct, we can merge.

I can confirm the parameters, these configs have already been used with real robots.

gavanderhoorn:
Closing and re-opening to nudge Actions.

gavanderhoorn:
And a comment: should the package manifests of both packages not be updated to state the dependency on the OPW Kinematics MoveIt plugin now, instead of trac ik?

gavanderhoorn:
thanks for keeping the `j23_coupled` commits out of this. It would be unrelated.

cjue:
The irb1200 config are IMHO already correct regarding `j23_coupled`. They have the opposite setting as compared to the irb1600 support packages.

cjue:
I rebased this to the current `kinetic-devel`

gavanderhoorn (approved):
Thanks for the PR :+1:

PR #139, opened 2022-06-29
Disable j23_coupled for robots without parallel arm linkage

cjue:
This PR builds upon #137.

Without this commit some support packages have J23_coupled enabled, some disabled.

To my knowledge however this would be an application specific setting that should match the robot controller configuration, and should rather be uniform between the support packages to avoid confusion.

`J23_coupled = false` matches the expected behavior for a generic 6-axis robot.

Do I misunderstand something that speaks against setting this to false in general?

### DIFF
--- a/abb_irb1600_support/launch/robot_interface_download_irb1600_6_12.launch
+++ b/abb_irb1600_support/launch/robot_interface_download_irb1600_6_12.launch
@@ -2,14 +2,14 @@
   Manipulator specific version of abb_driver's 'robot_interface.launch'.
 
   Defaults provided for IRB 1600:
-   - J23_coupled = true
+   - J23_coupled = false
 
   Usage:
     robot_interface_download_irb1600.launch robot_ip:=<value>
 -->
 <launch>
   <arg name="robot_ip" doc="IP of the controller" />
-  <arg name="J23_coupled" default="true" doc="If true, compensate for J2-J3 parallel linkage" />
+  <arg name="J23_coupled" default="false" doc="If true, compensate for J2-J3 parallel linkage" />
 
   <rosparam command="load" file="$(find abb_irb1600_support)/config/joint_names_irb1600_6_12.yaml" />
 
--- a/abb_irb1600_support/launch/robot_interface_download_irb1600_8_145.launch
+++ b/abb_irb1600_support/launch/robot_interface_download_irb1600_8_145.launch
@@ -2,14 +2,14 @@
   Manipulator specific version of abb_driver's 'robot_interface.launch'.
 
   Defaults provided for IRB 1600:
-   - J23_coupled = true
+   - J23_coupled = false
 
   Usage:
     robot_interface_download_irb1600.launch robot_ip:=<value>
 -->
 <launch>
   <arg name="robot_ip" doc="IP of the controller" />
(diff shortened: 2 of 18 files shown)


gavanderhoorn:
> Do I misunderstand something that speaks against setting this to false in general?

in my experience, the coupling is not application specific, but robot model/variant specific.

I've worked with controllers which had it set for their robot(s), and with controllers without it.

I trust the values we currently have set as `default` are correct. If they aren't, they should be fixed.

Without some documentation to back up your claim it's *application specific*, I'm hesitant to change a `default` in the way you propose.

gavanderhoorn:
Friendly ping here as well @cjue.

cjue:
> > Do I misunderstand something that speaks against setting this to false in general?
> 
> in my experience, the coupling is not application specific, but robot model/variant specific.

You are correct, I was confused by misreading part of the datasheet and the mismatch between the previous irb1600 default of `J23_coupled = true`. The setting is not application specific.

To my understanding models with external parallel arms like the IRB 6660 [1] should have the J23_coupled setting enabled, and robots without the external structure like the IRB 1200 [2] and 1600 [3] should have it disabled.

As I see it a lot of the robots in this repo have the wrong setting, for example the IRB1600.
We had to set the J23_coupled setting to false to be able to correctly operate the IRB1600 we were working on.

Overview of models: https://new.abb.com/products/robotics/industrial-robots/irb-1600

* [1] https://new.abb.com/products/3HAC020536-013/irb-6660
* [2] https://new.abb.com/products/robotics/industrial-robots/irb-1200
* [3] https://new.abb.com/products/robotics/industrial-robots/irb-1600

cjue:
I went over the list of all robots supported by abb_experimental and cross-checked their J23_coupled settings with their design documents.

If I did not miss something models with parallel arm linkage would be irb2400, 4400, 7700, 8700, 6660, 260, 460, 660, 760.
These all don't have support packages in `abb_experimental`.

The models irb1600, 6650s, 6700, 7600 and irb52 all don't have parallel arm linkage and therefore should have the J23_coupled setting disabled like the remaining models in this repository.

It appears to me, that the mistake arose when the IRB2400 model from the `abb` repo was used as a template for all following launch files, and that model does have J23_coupled correctly enabled.

cjue:
Note: if #137 is merged first, the irb1600 10/1.45 also needs to be modified.

gavanderhoorn:
I've finally had time to look into this (seems there aren't any other maintainers anymore).

Turns out: you're correct. I'm assuming people found out pretty quickly they had to change the value of `J23_coupled` if the default didn't work for them, otherwise I would've expected reports about this much sooner -- and more frequently.

Thank you for checking this and for submitting the PR.

We'll obviously have to fix `abb_irb2400_support` and `abb_irb4400_support` as well of course, over in `ros-industrial/abb`.

> It appears to me, that the mistake arose when the IRB2400 model from the `abb` repo was used as a template for all following launch files, and that model does have J23_coupled correctly enabled.

yeah, that seems like a good assumption. There's even the copy-pasta `2400` reference in one of the `1600` launch files.

gavanderhoorn:
> Note: if #137 is merged first, the irb1600 10/1.45 also needs to be modified.

I'll rebase and update #154 to include the change for that variant as well.

gavanderhoorn:
Closing in favour of #155.

Thanks for this PR @cjue :+1:

PR #147, opened 2023-10-19
Add irb2600 185m

aleunissen:

PR #156, opened 2025-04-05
Add support for 12/1.85 variant (rebased and cleaned #147)

gavanderhoorn:
As per title.

Cleaned up and rebased version of #147.

Squashed all commits in the first one, as they were all fixups.

Converted meshes to binary STL (from ASCII), corrected mesh origins and recreated the collision mesh.

Commit provenance and attribution retained.


gavanderhoorn:
As I've already reviewed #147, I'm not going to wait for a review here.

gavanderhoorn:
Thanks again @aleunissen and @Mennosytsma :+1:

